PR #1, opened 2016-03-27
Prevent out of bounds panic when value size exceeds c.config.MaxEntrySize

Freeaqingme:
```
panic: runtime error: slice bounds out of range

goroutine 17 [running, locked to thread]:
github.com/allegro/bigcache/queue.(*BytesQueue).allocateAdditionalMemory(0xc8200b6a08)
	/home/dolf/Projects/ClueGetter/src/github.com/allegro/bigcache/queue/bytes_queue.go:70 +0x512
github.com/allegro/bigcache/queue.(*BytesQueue).Push(0xc8200b6a08, 0xc821871300, 0x3b, 0x3b, 0xd)
	/home/dolf/Projects/ClueGetter/src/github.com/allegro/bigcache/queue/bytes_queue.go:54 +0x407
github.com/allegro/bigcache.(*BigCache).Set(0xc820072930, 0xc3abe0, 0xd, 0xc8217f56a0, 0x1c, 0x20)
	/home/dolf/Projects/ClueGetter/src/github.com/allegro/bigcache/bigcache.go:105 +0xb12
```

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -3,6 +3,7 @@ package bigcache
 import (
 	"log"
 	"sync"
+	"fmt"
 
 	"github.com/allegro/bigcache/queue"
 )
@@ -71,7 +72,7 @@ func (c *BigCache) Get(key string) ([]byte, error) {
 	}
 	if entryKey := readKeyFromEntry(wrappedEntry); key != entryKey {
 		if c.config.Verbose {
-			log.Printf("Collision detected. Both %q and %q has same hash %x", key, entryKey, hashedKey)
+			log.Printf("Collision detected. Both %q and %q have the same hash %x", key, entryKey, hashedKey)
 		}
 		return nil, notFound(key)
 	}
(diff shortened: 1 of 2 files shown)


Freeaqingme:
Coming to think of it, I'm not sure about this PR. Obviously a panic is bad, but the solution may not be ideal either. In my case I've got objects that can be as large as 48 MiB. However, given that most items will only be a few kb I'm not sure I should actually set the MaxEntrySize to 48 MiB.

druminski:
Hi @Freeaqingme, 

Thank you for the first open PR in this project. Currently there are no limits on value size and I prefer to keep it in that way. You should be able to put whatever you like to the cache (MaxEntrySize is only used to calculate initial size of the cache) Therefore what you have described is a bug. I have submitted fix for it: #2 Please, verify if it solves your problem. If it does then I will merge it to the master.

Freeaqingme:
@druminski Tnx! I don't seem to be able to trigger this panic anymore, so I'd say it's good to go.

=> Closed

janisz:
@druminski Consider renaming `MaxEntrySize` to `InitialCacheSize`

druminski:
@Freeaqingme awesome, merged.

@janisz ok, `InitialCacheSize` does not explain that this need to be size of an entry. I have added comment to that field, so I hope it is clearer.

PR #2, opened 2016-03-29
Handling values bigger than whole slice in bytes queue

druminski:
This PR fixes bug described in #1. It allows to set values bigger than currently allocated memory for cache.


wendigo:
:+1:

janisz:
:+1:

PR #5, opened 2016-03-29
Little optimizations

wendigo:
Hi,

I've applied two optimizations to bigcache:

- provided ability to use other hash functions like [farmhash](http://www.infoq.com/news/2014/04/google_farmhash) instead of fnv
- ensured that number of shards is power of two thus eliminating modulo operation in favor of bit mask

Results aggregated across 10 runs (before optimizations - with changes to number of shards in test cases):

```
BenchmarkWriteToCacheWith1Shard-8                           	1088 ns/op
BenchmarkWriteToCacheWith512Shards-8                        	342 ns/op
BenchmarkWriteToCacheWith1024Shards-8                       	342 ns/op
BenchmarkWriteToCacheWith8192Shards-8                       	359 ns/op
BenchmarkWriteToCacheWith1024ShardsAndSmallShardInitSize-8  	427 ns/op
BenchmarkReadFromCacheWith1024Shards-8                      	207 ns/op
BenchmarkReadFromCacheWith8192kShards-8                     	206 ns/op
```

Results aggregated across 10 runs (after optimizations):
```
BenchmarkWriteToCacheWith1Shard-8                           	1006 ns/op
BenchmarkWriteToCacheWith512Shards-8                        	331 ns/op
BenchmarkWriteToCacheWith1024Shards-8                       	316 ns/op
BenchmarkWriteToCacheWith8192Shards-8                       	326 ns/op
BenchmarkWriteToCacheWith1024ShardsAndSmallShardInitSize-8  	393 ns/op
BenchmarkReadFromCacheWith1024Shards-8                      	207 ns/op
BenchmarkReadFromCacheWith8192Shards-8                      	203 ns/op
```

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -1,6 +1,7 @@
 package bigcache
 
 import (
+	"fmt"
 	"log"
 	"sync"
 
@@ -18,8 +19,9 @@ type BigCache struct {
 	shards     []*cacheShard
 	lifeWindow uint64
 	clock      clock
-	hash       hash
+	hash       Hasher
 	config     Config
+	shardMask  uint64
 }
 
 type cacheShard struct {
@@ -30,17 +32,27 @@ type cacheShard struct {
 }
 
 // NewBigCache initialize new instance of BigCache
-func NewBigCache(config Config) *BigCache {
+func NewBigCache(config Config) (*BigCache, error) {
 	return newBigCache(config, &systemClock{})
 }
 
-func newBigCache(config Config, clock clock) *BigCache {
+func newBigCache(config Config, clock clock) (*BigCache, error) {
+
+	if !isPowerOfTwo(config.Shards) {
+		return nil, fmt.Errorf("Shards number must be power of two")
+	}
+
+	if config.Hasher == nil {
+		config.Hasher = newDefaultHasher()
+	}
+
 	cache := &BigCache{
 		shards:     make([]*cacheShard, config.Shards),
 		lifeWindow: uint64(config.LifeWindow.Seconds()),
 		clock:      clock,
-		hash:       fnv64a{},
+		hash:       config.Hasher,
 		config:     config,
+		shardMask:  uint64(config.Shards - 1),
 	}
 
 	shardSize := max(config.MaxEntriesInWindow/config.Shards, minimumEntriesInShard)
@@ -51,12 +63,17 @@ func newBigCache(config Config, clock clock) *BigCache {
 			entryBuffer: make([]byte, config.MaxEntrySize+headersSizeInBytes),
 		}
 	}
-	return cache
+
+	return cache, nil
+}
+
+func isPowerOfTwo(number int) bool {
+	return (number & (number - 1)) == 0
 }
 
 // Get reads entry for the key
 func (c *BigCache) Get(key string) ([]byte, error) {
-	hashedKey := c.hash.sum(key)
+	hashedKey := c.hash.Sum64(key)
 	shard := c.getShard(hashedKey)
 	shard.lock.RLock()
 	defer shard.lock.RUnlock()
@@ -82,7 +99,7 @@ func (c *BigCache) Get(key string) ([]byte, error) {
 
 // Set saves entry under the key
 func (c *BigCache) Set(key string, entry []byte) {
-	hashedKey := c.hash.sum(key)
+	hashedKey := c.hash.Sum64(key)
 	shard := c.getShard(hashedKey)
 	shard.lock.Lock()
 	defer shard.lock.Unlock()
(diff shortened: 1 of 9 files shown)


hash.go
```diff
@@ -1,16 +1,20 @@
 package bigcache
 
-import "hash/fnv"
+import farm "github.com/dgryski/go-farm"
```
janisz: I'm not fan of adding external dependency to this project

bigcache.go
```diff
@@ -29,17 +32,23 @@ type cacheShard struct {
...
+	if (!isPowerOfTwo(config.Shards)) {
+		return nil, fmt.Errorf("Shards number must be power of two")
+	}
+
 	cache := &BigCache{
 		shards:     make([]*cacheShard, config.Shards),
 		lifeWindow: uint64(config.LifeWindow.Seconds()),
 		clock:      clock,
-		hash:       fnv64a{},
+		hash:       newHasher(),
```
janisz: Maybe we should move hash function to config and if it's not set use `fnv64a`

bigcache.go
```diff
@@ -49,7 +58,12 @@ func newBigCache(config Config, clock clock) *BigCache {
 			entries: *queue.NewBytesQueue(shardSize*config.MaxEntrySize, config.Verbose),
 		}
 	}
-	return cache
+
+	return cache, nil
+}
+
+func isPowerOfTwo(number int) bool {
+	return number == int(math.Pow(2, math.Ceil(math.Log2(float64(number)))))
```
janisz: Should be [`return ((number & (number - 1)) == 0)`](http://bits.stephan-brumme.com/isPowerOfTwo.html)

bigcache.go
```diff
@@ -114,8 +128,7 @@ func (c *BigCache) onEvict(oldestEntry []byte, currentTimestamp uint64, evict fu
 }
 
 func (c *BigCache) getShard(hashedKey uint64) (shard *cacheShard) {
-	shardKey := hashedKey % uint64(len(c.shards))
-	return c.shards[shardKey]
+	return c.shards[hashedKey & c.shardMask]
```
janisz: :+1: for using bit mask instead of modulo

druminski: I also like that optimization

hash.go
druminski: If we want use it then it would be nice to have hash bench test proving that go-farm is faster than fnv.

wendigo:
As for now I introduced hasher in the config + provided defaults

hash.go
```diff
@@ -1,15 +1,21 @@
 package bigcache
 
-import "hash/fnv"
+import (
+	"hash/fnv"
+)
 
-type hash interface {
-	sum(string) uint64
+type Hasher interface {
```
janisz: > exported type Hasher should have comment or be unexported

bigcache.go
```diff
@@ -29,17 +32,27 @@ type cacheShard struct {
...
 // NewBigCache initialize new instance of BigCache
-func NewBigCache(config Config) *BigCache {
+func NewBigCache(config Config) (*BigCache, error) {
 	return newBigCache(config, &systemClock{})
 }
 
-func newBigCache(config Config, clock clock) *BigCache {
+func newBigCache(config Config, clock clock) (*BigCache, error) {
+
+	if !isPowerOfTwo(config.Shards) {
```
janisz: Add note to `Config.Shards` about power of two requirement

bigcache.go
```diff
@@ -4,6 +4,8 @@ import (
 	"log"
 	"sync"
```
janisz: Delete this line and format imports

hash.go
wendigo: ok

bigcache.go
wendigo: ok

bigcache.go
wendigo: ok

encoding.go
```diff
@@ -14,14 +14,16 @@ const (
...
+
+	blob = make([]byte, totalLength)
 	binary.LittleEndian.PutUint64(blob, timestamp)
 	binary.LittleEndian.PutUint64(blob[timestampSizeInBytes:], hash)
 	binary.LittleEndian.PutUint16(blob[timestampSizeInBytes+hashSizeInBytes:], uint16(keyLength))
 	copy(blob[headersSizeInBytes:], []byte(key))
 	copy(blob[headersSizeInBytes+keyLength:], entry)
 
-	return blob
+	return blob[:totalLength]
```
janisz: Revert changes in this file. They will conflict with #7

druminski:
:+1:

PR #7, opened 2016-03-31
Wrap entry optimization

druminski:
Alternative implementation of wrap entry optimization proposed by @wendigo in #5 

Instead of using global sync.Pool with locks, entry byte buffer is allocated per shard.

Bench test results before and after optimization:

```
go test -bench=WriteToCacheWith -benchtime=5s

BenchmarkWriteToCacheWith1Shard-4                       5000000	     1520 ns/op
BenchmarkWriteToCacheWith500Shards-4                    	10000000	      647 ns/op
BenchmarkWriteToCacheWith1kShards-4                     10000000	      627 ns/op
BenchmarkWriteToCacheWith10kShards-4                    	10000000	      646 ns/op
BenchmarkWriteToCacheWith1kShardsAndSmallShardInitSize-4	10000000	      789 ns/op

BenchmarkWriteToCacheWith1Shard-4                       5000000	     1320 ns/op
BenchmarkWriteToCacheWith500Shards-4                    	20000000	      594 ns/op
BenchmarkWriteToCacheWith1kShards-4                     20000000	      616 ns/op
BenchmarkWriteToCacheWith10kShards-4                    	10000000	      629 ns/op
BenchmarkWriteToCacheWith1kShardsAndSmallShardInitSize-4	10000000	      717 ns/op
```

Total time spent on entries wrapping during bench tests:

```
//before opt
1.13s     15.34s    104:	w := wrapEntry(currentTimestamp, hashedKey, key, entry)

//after opt
2.50s      3.84s    106:	w := wrapEntry(currentTimestamp, hashedKey, key, entry, &shard.entryBuffer)
```

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -11,17 +11,22 @@ const (
 	headersSizeInBytes   = timestampSizeInBytes + hashSizeInBytes + keySizeInBytes // Number of bytes used for all headers
 )
 
-func wrapEntry(timestamp uint64, hash uint64, key string, entry []byte) []byte {
-	var blob []byte
+func wrapEntry(timestamp uint64, hash uint64, key string, entry []byte, buffer *[]byte) []byte {
 	keyLength := len(key)
-	blob = make([]byte, len(entry)+headersSizeInBytes+keyLength)
+	blobLength := len(entry) + headersSizeInBytes + keyLength
+
+	if blobLength > len(*buffer) {
+		*buffer = make([]byte, blobLength)
+	}
+	blob := *buffer
+
 	binary.LittleEndian.PutUint64(blob, timestamp)
 	binary.LittleEndian.PutUint64(blob[timestampSizeInBytes:], hash)
 	binary.LittleEndian.PutUint16(blob[timestampSizeInBytes+hashSizeInBytes:], uint16(keyLength))
 	copy(blob[headersSizeInBytes:], []byte(key))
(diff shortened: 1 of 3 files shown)


wendigo:
Wow, nice one :) It seems that using sync.Pool has no sense here and this kind of buffer is fine in wrapEntry

druminski:
I am glad you like it, although credits goes to you with this pr as you discovered the optimization possibility and proposed a solution with comparable results, thanks :)

wendigo:
:+1:

encoding.go
```diff
@@ -11,17 +11,22 @@ const (
 	headersSizeInBytes   = timestampSizeInBytes + hashSizeInBytes + keySizeInBytes // Number of bytes used for all headers
 )
 
-func wrapEntry(timestamp uint64, hash uint64, key string, entry []byte) []byte {
-	var blob []byte
+func wrapEntry(timestamp uint64, hash uint64, key string, entry []byte, buffer *[]byte) []byte {
```
janisz: 5 arguments is quite much. How about introducing
```
type Entry struct {
    timestamp uint64
    hash uint64
    key uint64
    entry []byte
}
```

janisz:
I leave one comment but it looks good. :+1:

encoding.go
druminski: I think it's a little bit to soon for this structure, I will happily use it if it will be used more than in one place to pass its values to func.

PR #8, opened 2016-04-03
Entries indexes stay unchanged after additional memory allocation

druminski:
When additional memory allocation occurs in bytes queue then its entries indexes should stay unchanged, because  values from bigcache map can point to them. Proposed fix fills space between tail and head pointers with an empty blob, unit tests show how bytes queue behaves in that situation.

### DIFF
--- a/queue/bytes_queue_test.go
+++ b/queue/bytes_queue_test.go
@@ -40,17 +40,17 @@ func TestReuseAvailableSpace(t *testing.T) {
 	t.Parallel()
 
 	// given
-	queue := NewBytesQueue(21, false)
+	queue := NewBytesQueue(100, false)
 
 	// when
-	queue.Push([]byte("hello1"))
-	queue.Push([]byte("hello2"))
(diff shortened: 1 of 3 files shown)


queue/bytes_queue_test.go
```diff
@@ -40,17 +40,17 @@ func TestReuseAvailableSpace(t *testing.T) {
...
+	queue.Push(blob('c', 20))
 
 	// then
-	assert.Equal(t, 21, queue.Capacity())
-	assert.Equal(t, []byte("hello2"), pop(queue))
+	assert.Equal(t, 100, queue.Capacity())
+	assert.Equal(t, blob('b', 20), pop(queue))
 }
 
 func TestAllocateAdditionalSpace(t *testing.T) {
```
janisz: What do you think about leaving old test untouched (happy path) and add new with more descriptive name (special case)

janisz:
Good catch and great patch :+1:

PR #10, opened 2016-04-12
Implementing hard max cache size

druminski:


README.md
```diff
@@ -37,6 +37,7 @@ config := bigcache.Config{
 		MaxEntriesInWindow: 1000 * 10 * 60, // rps * lifeWindow
 		MaxEntrySize: 500,                  // max entry size in bytes, used only in initial memory allocation
 		Verbose: true,                      // prints information about additional memory allocation
+		HardMaxCacheSize: 8192,             // cache will not allocate more memory than this limit, value in MB, 0 means no limit
```
janisz: If cache will not allocate more memory than limit what will happen? It should be described here. I think `Hard` prefix is not needed here

bigcache.go
```diff
@@ -46,20 +47,26 @@ func newBigCache(config Config, clock clock) (*BigCache, error) {
 		config.Hasher = newDefaultHasher()
 	}
 
+	maxShardSize := 0
+	if config.HardMaxCacheSize > 0 {
+		maxShardSize = config.HardMaxCacheSize * 1024 * 1024 / config.Shards
```
janisz: Magic numbers? It's not obvious why you are multiplying by 1024. Maybe add unit to `maxHardSize`

druminski: ok, extracted to convertMBToBytes()

README.md
druminski: It is described in config.go but ok I have added also info here. As for `Hard` prefix I decided to leave it because we already have `MaxEntriesInWindow` and `MaxEntrySize` which are soft limits - used only in initial memory allocation, therefore I wanted to distinct them.

bigcache.go
```diff
@@ -46,20 +47,26 @@ func newBigCache(config Config, clock clock) (*BigCache, error) {
 		config.Hasher = newDefaultHasher()
 	}
 
+	maxShardSize := 0
+	if config.HardMaxCacheSize > 0 {
+		maxShardSize = convertMBToBytes(config.HardMaxCacheSize) / config.Shards
```
janisz: So it is possible to overwrite data even if total size of cache will be smaller than limit.

janisz:
LGMT

Freeaqingme:
Sweet! I'll probably let you know something over the weekend.

druminski:
No further comments, merging

PR #11, opened 2016-04-13
Add some tests

janisz:
Use `DefaultConfig` in test
Make queue tests logs verbose
Add somne corner case tests

### DIFF
--- a/queue/bytes_queue_test.go
+++ b/queue/bytes_queue_test.go
@@ -10,8 +10,14 @@ func TestPushAndPop(t *testing.T) {
 	t.Parallel()
 
 	// given
-	queue := NewBytesQueue(10, true)
+	queue := NewBytesQueue(10, false)
 	entry := []byte("hello")
+
+	// when
+	_, err := queue.Pop()
+
+	// then
+	assert.EqualError(t, err, "Empty queue")
 
 	// when
 	queue.Push(entry)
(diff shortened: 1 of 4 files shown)


queue/bytes_queue_test.go
```diff
@@ -226,6 +255,30 @@ func TestGetEntryFromInvalidIndex(t *testing.T) {
...
+
+	// given
+	queue := NewBytesQueue(13, true)
+
+	// when
+	result, err := queue.Get(1)
+
+	// then
+	assert.NotNil(t, result)
+	assert.NoError(t, err)
```
janisz: Maybe we should return Empty queue error here.

queue/bytes_queue_test.go
```diff
@@ -226,6 +255,30 @@ func TestGetEntryFromInvalidIndex(t *testing.T) {
...
 }
 
+func TestGetEntryFromIndexOutOfRange(t *testing.T) {
+	t.Parallel()
+
+	// given
+	queue := NewBytesQueue(0, true)
+
+	// when
+	assert.Panics(t, func() { queue.Get(42) })
```
janisz: Maybe we should return error instead of Panicking.

druminski: sure, we can add some args validation in the constructor. Also args in newBigCache constructor also can be validated, for example MaxEntriesInWindow or MaxEntrySize need to be greater than 0.

queue/bytes_queue_test.go
druminski: good remark, it will be a better behavior in that situation

queue/bytes_queue_test.go
```diff
@@ -71,7 +100,7 @@ func TestAllocateAdditionalSpaceForInsufficientFreeFragmentedSpaceWhereHeadIsBef
 	t.Parallel()
 
 	// given
-	queue := NewBytesQueue(25, false)
+	queue := NewBytesQueue(25, true)
```
druminski: Output from bench tests is cleaner when verbose mode in unit tests is disabled as bench tests also run unit tests.

queue/bytes_queue_test.go
janisz: Fixed

queue/bytes_queue_test.go
janisz: Fixed

queue/bytes_queue_test.go
janisz: Fixed

druminski:
:+1:

PR #15, opened 2016-05-06
resolves #14 | Adding OnRemove handler

### LINKED ISSUE #14: Event handlers for evict?
Is it possible to have eviction handler like in the lib https://github.com/bluele/gcache? Because we need to do some cleanup whenever a cache key is evicted.

I checked out your source and it seems you do have some support for it, buts internal only here
https://github.com/allegro/bigcache/blob/master/bigcache.go#L138

druminski:


janisz:
LGTM

PR #17, opened 2016-05-09
Build only for go 1.6

janisz:
Because our builds starts to fail due to problems with third party tools on go 1.7 we want to run build only on stable releases.


druminski:
👍

PR #22, opened 2016-10-12
Add Iterator to iterate over keys

wendigo:


bigcache.go
```diff
@@ -33,6 +33,19 @@ type cacheShard struct {
...
+	Key       string
+	Hash      uint64
+	Timestamp uint64
+}
+
+// EntryInfoFilter function allows filtering of EntryInfo elements in the KeysIterator functions
+type EntryInfoFilter func(EntryInfo) bool
+
+// EntryInfoIterator allows to iterate over entries in the cache
+type EntryInfoIterator chan EntryInfo
```
janisz: How about using closure instead?

janisz: With channel you add blocking. Take a look at loops you created. You are adding mutex for reading from map and then pushing value to channel which is blocking. If you remove channel and instead return closure to function that returns entry form shard `i` on position `index` then you don't need to copy all (1024) elements and this will be a lazy iterator.

wendigo: Channel is buffered

bigcache.go
```diff
@@ -33,6 +33,79 @@ type cacheShard struct {
...
+	Timestamp uint64
+	index     uint32
+}
+
+// EntryInfoIterator allows to iterate over entries in the cache
+type EntryInfoIterator struct {
+	cache        *BigCache
+	currentShard int
+	currentIndex int32
+	elements     []*EntryInfo
```
janisz: This could be just slice of indexes `[]uin32`. We don't need all data during iteration.

bigcache.go
```diff
@@ -33,6 +33,79 @@ type cacheShard struct {
...
+}
+
+func copyCurrentShardMap(shard *cacheShard) []*EntryInfo {
+	shard.lock.RLock()
+	defer shard.lock.RUnlock()
+
+	elements := make([]*EntryInfo, 0)
```
janisz: Try to estimate initial size for elements as `len(shard.queue)/cache.config.MaxEntrySize`

bigcache_test.go
```diff
@@ -113,6 +115,36 @@ func TestEntryUpdate(t *testing.T) {
...
+	}
+
+	// when
+	keys := make(map[string]struct{})
+	iterator := cache.Iterator()
+
+	for iterator.Next() {
+		current, err := iterator.Value()
+
+		if err == nil {
```
janisz: Add test when error occures: modify map when it terates, delete element under `Value()`

bigcache.go
```diff
@@ -33,6 +33,79 @@ type cacheShard struct {
...
+	}
+}
+
+// Value returns current value from the iterator
+func (it *EntryInfoIterator) Value() (*EntryInfo, error) {
+	current := it.elements[it.currentIndex]
+
+	if entry, err := it.cache.shards[it.currentShard].entries.Get(int(current.index)); err != nil {
+		return nil, fmt.Errorf("Could not retrieve entry from cache")
+	} else {
```
janisz: > if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (golint)

bigcache.go
```diff
@@ -33,6 +33,79 @@ type cacheShard struct {
 	onRemove    func(wrappedEntry []byte)
 }
 
+// EntryInfo holds informations about entry in the cache
+type EntryInfo struct {
```
janisz: Why not contain entry value?

bigcache_test.go
janisz: It also needs bench test.

wendigo: Benchmark - done.

bigcache.go
wendigo: done

bigcache.go
```diff
@@ -33,6 +33,111 @@ type cacheShard struct {
...
+
+		if it.currentShard == it.cache.config.Shards {
+			return false
+		}
+
+		it.elements = copyCurrentShardMap(it.cache.shards[it.currentShard])
+
+		// partition is empty - move to next
+		if len(it.elements) == 0 {
+			goto incrementShard
```
janisz: Just use `for` instead of `if` above the label

wendigo: Why? :) I don't understand the pros

janisz: I mean change if in line 90 (above the label)  and remove this condition. You are implementing `for` loop with `if` and `goto `

wendigo: Yep, that's exactly what I'm doing and code is readable for me. Goto is a standard jmp in the code

wendigo: Nevermind - I rewrite this part of iterator

bigcache.go
wendigo: Reason is simple: entryinfo is lazy - you will only parse what you need instead of everything

bigcache.go
janisz: I'd prefer to use `for` loop and extract condition to well named function so the comment will be redundant.

bigcache.go
```diff
@@ -33,6 +33,110 @@ type cacheShard struct {
...
+		currentIndex: -1,
+		elements:     copyCurrentShardMap(cache.shards[0]),
+	}
+}
+
+// Value returns current value from the iterator
+func (it *EntryInfoIterator) Value() (EntryInfo, error) {
+	it.Lock()
+	defer it.Unlock()
+	current := it.elements[it.currentIndex]
```
janisz: I think we should check  if current index is valid.

bigcache.go
```diff
@@ -33,6 +33,110 @@ type cacheShard struct {
...
+	if it.currentIndex >= len(it.elements) {
+
+		// Last shard - no more entries
+		if it.currentShard == it.cache.config.Shards-1 {
+			return false
+		}
+
+		for i := it.currentShard + 1; i < it.cache.config.Shards; i++ {
+			it.currentShard, it.currentIndex, it.elements = i, 0, copyCurrentShardMap(it.cache.shards[i])
+
```
janisz: This line is too long.

bigcache.go
```diff
@@ -33,6 +33,110 @@ type cacheShard struct {
...
+}
+
+// Next returns true if there is next element in the iterator
+func (it *EntryInfoIterator) Next() bool {
+	it.Lock()
+	defer it.Unlock()
+
+	it.currentIndex++
+
+	if it.currentIndex >= len(it.elements) {
```
janisz: How about reverting condition. Just for readability.

bigcache.go
```diff
@@ -33,6 +33,110 @@ type cacheShard struct {
...
+
+		// Last shard - no more entries
+		if it.currentShard == it.cache.config.Shards-1 {
+			return false
+		}
+
+		for i := it.currentShard + 1; i < it.cache.config.Shards; i++ {
```
janisz: What if all shards are empty?

bigcache.go
```diff
@@ -33,6 +33,110 @@ type cacheShard struct {
...
+	var elements []uint32
+
+	for _, index := range shard.hashmap {
+		elements = append(elements, index)
+	}
+
+	return elements
+}
+
+// Next returns true if there is next element in the iterator
```
janisz: This describes only half of the job that `Next() ` do becouse it also moves iterator to next element or invalid position if it doesn't exist. This comment better suits `HasNext() `

bigcache.go
wendigo: Ok

bigcache.go
wendigo: You mean: if !(len(it.elements) > it.currentIndex) ?

bigcache.go
```diff
@@ -33,6 +33,121 @@ type cacheShard struct {
...
+
+// HasNext returns true if there is next element in the iterator, false otherwise
+func (it *EntryInfoIterator) HasNext() bool {
+	it.Lock()
+	defer it.Unlock()
+
+	it.valid = false
+	it.currentIndex++
+
+	if !(len(it.elements) > it.currentIndex) {
```
janisz: Oh, I was thinking about early return `true` so the code below doesn't need to be indent.

```
if len(it.elements) > it.currentIndex {
   it.valid = true;
    return true
}
```

bigcache.go
janisz: My mistake, looks good.

bigcache_test.go
```diff
@@ -113,6 +114,121 @@ func TestEntryUpdate(t *testing.T) {
...
+	}
+
+	current, err := iterator.Value()
+
+	// then
+	assert.Nil(t, err)
+	assert.Equal(t, current.Key(), "key")
+	assert.Equal(t, current.Hash(), uint64(0x3dc94a19365b10ec))
+	assert.Equal(t, current.Value(), []byte("value"))
+	assert.Equal(t, current.Timestamp(), uint64(0))
```
janisz: Parameters order mismatch [`func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool`](https://godoc.org/github.com/stretchr/testify/assert#Equal). Check with other places too.

bigcache.go
janisz: So whats whit this? I did a small benchmark and gready extraction is a little bit slower than copy.
<details>
```diff
diff --git a/bigcache.go b/bigcache.go
index d610216..f15c1de 100644
--- a/bigcache.go
+++ b/bigcache.go
@@ -35,27 +35,30 @@ type cacheShard struct {
 
 // EntryInfo holds informations about entry in the cache
 type EntryInfo struct {
-       entry []byte
+       timestamp uint64
+       hash      uint64
+       key       string
+       value     []byte
 }
 
 // Key returns entry's underlying key
 func (e EntryInfo) Key() string {
-       return readKeyFromEntry(e.entry)
+       return e.key
 }
 
 // Hash returns entry's hash value
 func (e EntryInfo) Hash() uint64 {
-       return readHashFromEntry(e.entry)
+       return e.hash
 }
 
 // Timestamp returns entry's timestamp (time of insertion)
 func (e EntryInfo) Timestamp() uint64 {
-       return readTimestampFromEntry(e.entry)
+       return e.timestamp
 }
 
 // Value returns entry's underlying value
 func (e EntryInfo) Value() []byte {
-       return readEntry(e.entry)
+       return e.value
 }
 
 // EntryInfoIterator allows to iterate over entries in the cache
@@ -140,11 +143,11 @@ func (it *EntryInfoIterator) Value() (EntryInfo, error) {
                return EntryInfo{}, fmt.Errorf("Could not retrieve entry from cache")
        }
 
-       var dst = make([]byte, len(entry))
-       copy(dst, entry)
-
        return EntryInfo{
-               entry: dst,
+               timestamp: readTimestampFromEntry(entry),
+               hash:      readHashFromEntry(entry),
+               key:       readKeyFromEntry(entry),
+               value:     readEntry(entry),
        }, nil
 }
```

```shell
$go test ./... -bench=BenchmarkIterateOverCache -run=NONE -count 3 -cpu 1,6 -benchmem > new.txt && git stash && go test ./... -bench=BenchmarkIterateOverCache -run=NONE -count 3 -cpu 1,6 benchmem > old.txt 
$ benchcmp old.txt new.txt                                                                                                                                                                                     
benchmark                                                    old ns/op     new ns/op     delta
BenchmarkIterateOverCache/512-shards        224           224           +0.00%
BenchmarkIterateOverCache/512-shards        217           223           +2.76%
BenchmarkIterateOverCache/512-shards        218           248           +13.76%
BenchmarkIterateOverCache/512-shards-6      273           293           +7.33%
BenchmarkIterateOverCache/512-shards-6      273           284           +4.03%
BenchmarkIterateOverCache/512-shards-6      274           284           +3.65%
BenchmarkIterateOverCache/1024-shards       214           216           +0.93%
BenchmarkIterateOverCache/1024-shards       214           215           +0.47%
BenchmarkIterateOverCache/1024-shards       215           218           +1.40%
BenchmarkIterateOverCache/1024-shards-6     268           279           +4.10%
BenchmarkIterateOverCache/1024-shards-6     271           278           +2.58%
BenchmarkIterateOverCache/1024-shards-6     268           281           +4.85%
BenchmarkIterateOverCache/8192-shards       209           210           +0.48%
BenchmarkIterateOverCache/8192-shards       207           210           +1.45%
BenchmarkIterateOverCache/8192-shards       207           211           +1.93%
BenchmarkIterateOverCache/8192-shards-6     265           276           +4.15%
BenchmarkIterateOverCache/8192-shards-6     266           275           +3.38%
BenchmarkIterateOverCache/8192-shards-6     263           279           +6.08%
```
</details>

bigcache.go
wendigo: Done :) Renamed ```HasNext``` to ```SetNext``` also

bigcache_test.go
wendigo: Done.

bigcache.go
```diff
@@ -33,6 +33,121 @@ type cacheShard struct {
...
+
+	for _, index := range shard.hashmap {
+		elements = append(elements, index)
+	}
+
+	return elements
+}
+
+// HasNext returns true if there is next element in the iterator, false otherwise
+func (it *EntryInfoIterator) HasNext() bool {
```
wendigo: You're right - renamed method to SetNext().

wendigo:
Waiting for final thoughts @janisz

bigcache.go
```diff
@@ -12,6 +13,12 @@ const (
 	minimumEntriesInShard = 10 // Minimum number of entries in single shard
 )
 
+// ErrInvalidIteratorState is reported when iterator is in invalid state and cannot be used.
+var ErrInvalidIteratorState = errors.New("Iterator is in invalid state. Use SetNext() to move to next position.")
+
+// ErrCannotRetrieveEntry is reported when entry cannot be retrieved from underlying
+var ErrCannotRetrieveEntry = errors.New("Could not retrieve entry from cache")
+
```
janisz: This two should not be public. Use empty struct withe Error method like we do in [entry_not_found_error.go](https://github.com/allegro/bigcache/blob/master/entry_not_found_error.go) or make this private. Take a look at http://dave.cheney.net/2016/04/07/constant-errors

bigcache.go
janisz: What with this?

bigcache_bench_test.go
```diff
@@ -17,6 +17,8 @@ func BenchmarkWriteToCacheWith1Shard(b *testing.B) {
 func BenchmarkWriteToLimitedCacheWithSmallInitSizeAnd1Shard(b *testing.B) {
 	m := blob('a', 1024)
 	cache, _ := NewBigCache(Config{1, 100 * time.Second, 100, 256, false, nil, 1, nil})
+
+	b.ReportAllocs()
```
janisz: If applied to all tests then better use `-test.benchmem`

bigcache_test.go
```diff
@@ -113,6 +114,121 @@ func TestEntryUpdate(t *testing.T) {
 	assert.Equal(t, []byte("value2"), cachedValue)
 }
 
+func TestEntriesIterator(t *testing.T) {
+	t.Parallel()
+
+	// given
+	keysCount := 10000
```
janisz: Why do we need so many entries in unit test?

bigcache_bench_test.go
wendigo: It will include setup also (creating entries for reads etc)

bigcache.go
wendigo: What do you mean "What with this"?  :)

janisz: Not this line but I assume you are referring to this discussion https://github.com/allegro/bigcache/pull/22#discussion_r83423780

bigcache_test.go
wendigo: Reduced by 10x

bigcache.go
wendigo: Ok, done, it saves some allocs (20B/op vs 32B/op). Times are the a little smaller.

PR #24, opened 2016-11-02
Zero alloc byte to string conversion

wendigo:
No copying when converting []byte to string.

```
go test -bench BenchmarkIterateOverCache -memprofile mem.out
2016/11/03 00:29:24 Collision detected. Both "liquid" and "costarring" have the same hash 5
BenchmarkIterateOverCache/512-shards-8         	 5000000	       397 ns/op	       4 B/op	       0 allocs/op
BenchmarkIterateOverCache/1024-shards-8        	 5000000	       398 ns/op	       4 B/op	       0 allocs/op
BenchmarkIterateOverCache/8192-shards-8        	 5000000	       374 ns/op	       4 B/op	       0 allocs/op
```

Also we should use []byte keys instead of strings...

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -25,7 +25,7 @@ func wrapEntry(timestamp uint64, hash uint64, key string, entry []byte, buffer *
 	binary.LittleEndian.PutUint64(blob, timestamp)
 	binary.LittleEndian.PutUint64(blob[timestampSizeInBytes:], hash)
 	binary.LittleEndian.PutUint16(blob[timestampSizeInBytes+hashSizeInBytes:], uint16(keyLength))
-	copy(blob[headersSizeInBytes:], stringToBytes(key))
+	copy(blob[headersSizeInBytes:], key)
 	copy(blob[headersSizeInBytes+keyLength:], entry)
 
 	return blob[:blobLength]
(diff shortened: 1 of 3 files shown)


encoding.go
```diff
@@ -40,7 +42,13 @@ func readTimestampFromEntry(data []byte) uint64 {
 
 func readKeyFromEntry(data []byte) string {
 	length := binary.LittleEndian.Uint16(data[timestampSizeInBytes+hashSizeInBytes:])
-	return string(data[headersSizeInBytes : headersSizeInBytes+length])
+	return bytesToString(data[headersSizeInBytes : headersSizeInBytes+length])
+}
+
+func bytesToString(b []byte) string {
+	bytesHeader := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+	strHeader := reflect.StringHeader{Data: bytesHeader.Data, Len: bytesHeader.Len}
```
janisz: > Clear is better than clever.
> Reflection is never clear.

janisz: What will happen when I read key from the cache and then underlying value is overwritten? Will I have my value?

wendigo: Yep. Exactly the same as with ```readEntry()```

wendigo: Probably we should copy data in both ```readEntry()``` and ```readKeyFromEntry()```

janisz: Is this expected behavior? For example when you iterate over entries and keep entry info Key could change over time. I think we need to document this behavior and provide test as an example what can happen.

wendigo: Ok, I've added test that will fail if copy won't be made in encoding.go ```readEntry```

Issue #23, opened 2016-11-01
Clear Cache


btesanovic:
Hay there will it be possible to add clear cache method
Thanks

janisz:
I think we can add this method. But how it should works in details. I assume after calling `Clear()` we should end up with no entries in cache. Questions we need to answer

1. Are we going to call [`OnRemove` callback](https://github.com/allegro/bigcache/blob/master/config.go#L27) for every element in cache
2. Are we going to create new (small) shard queues or just use existing one

In my opinion clear should just delete all entries in maps by creating new map (GC will do the rest) and reset indexes in queue (`tail`, `head`, `rightMargin`, `count`). `OnRemove` should NOT be called.

@btesanovic @wendigo What do you think?

wendigo:
@janisz: ```onRemove``` does not make any sense I guess (if you need keys - use iterator before clear()). I agree with you on creating new map and reseting indexes

wendigo:
I'll try to come up with an implementation

wendigo:
Solved @btesanovic. Let me know if it works for you!

btesanovic:
Hey @wendigo  , thanks so much for fast response, this is perfect !!

PR #25, opened 2016-11-03
Add Reset() to cache

wendigo:
* Add ```Reset()``` method to ```BigCache``` (reset all shards)
* Add ```Reset()``` method to BytesQueue (reset queue to initial state)
* Add ```Len()``` method to ```BigCache``` (number of entries)
* Move cacheShard to it's own file
* Move config logic

Addresses https://github.com/allegro/bigcache/issues/23

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -1,9 +1,6 @@
 package bigcache
 
-import (
-	"errors"
-	"sync"
-)
+import "sync"
 
 type iteratorError string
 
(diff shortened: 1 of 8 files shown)


iterator.go
```diff
@@ -47,7 +44,7 @@ func (e EntryInfo) Value() []byte {
 
 // EntryInfoIterator allows to iterate over entries in the cache
 type EntryInfoIterator struct {
-	sync.Mutex
+	mutex         sync.Mutex
```
janisz: What's the reason to change it?

queue/bytes_queue.go
```diff
@@ -39,17 +40,28 @@ type queueError struct {
...
 	}
 }
 
+// Reset removes all entries from queue
+func (q *BytesQueue) Reset() {
+	// Just reset indexes
+	q.tail = leftMarginIndex
+	q.head = leftMarginIndex
+	q.rightMargin = leftMarginIndex
+	q.capacity = q.initialCapacity
```
janisz: Do we need to set it with reset? Capacity shouldn't be changed because we didn't shrink `q.array`

shard.go
```diff
@@ -0,0 +1,50 @@
+package bigcache
```
janisz: Nice move :)

iterator.go
wendigo: It was exposed as public Mutex field

PR #40, opened 2017-07-07
Cache Distribution

mxplusb:
Adds io.Writer interface to flush cache on eviction and/or manually via new bigcache.Flush() API to an arbitrary destination.

Tests to match.

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -82,6 +82,7 @@ func (c *BigCache) Get(key string) ([]byte, error) {
 		if c.config.Verbose {
 			log.Printf("Collision detected. Both %q and %q have the same hash %x", key, entryKey, hashedKey)
 		}
+		shard.lock.RUnlock()
 		return nil, notFound(key)
 	}
 	return readEntry(wrappedEntry), nil
@@ -92,7 +93,6 @@ func (c *BigCache) Set(key string, entry []byte) error {
 	hashedKey := c.hash.Sum64(key)
 	shard := c.getShard(hashedKey)
 	shard.lock.Lock()
-	defer shard.lock.Unlock()
 
 	currentTimestamp := uint64(c.clock.epoch())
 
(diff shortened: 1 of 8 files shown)


mxplusb:
So to use gob encoding in Go, you have to have a fixed-width data type, so no slices. Due to the sharding design, without doing an entire rewrite of the whole sharding subsystem I won't be able to use gob encoding for flushing data to a writer. I found this out once I started looking at where coveralls said the test coverage was lower. I followed a crazy rabbit hole of good and bad ideas, and decided `encoding/gob` wasn't he right solution.

Once gob was determined to not be the right solution (spent like two hours on it), JSON seemed to be the next appropriate encoder. In order to have the proper schema for exporting the cache, I had a couple options. I could either continue with unexported field off the primary struct types and implement a custom marshaler, which would copy the relevant data into a new unexported struct with exported fields, or I could export the necessary fields on the current structs. The latter seemed less invasive and indirect.

Now that this aspect is done, I'm going to work on cache importing via `io.Reader`. With both of those interfaces in place, you'd then be able to export the cache and/or import a pre-built cache without having to repopulate it.

This is useful for distributed cache design across multiple machines, especially if you have a cache cluster, with a master -> slaves relationship for HA/Failover designs.

bigcache.go
```diff
@@ -13,13 +14,13 @@ const (
...
+	Shards       []*cacheShard
+	LifeWindow   uint64
 	clock        clock
 	hash         Hasher
-	config       Config
-	shardMask    uint64
-	maxShardSize uint32
+	Config       Config
+	ShardMask    uint64
+	MaxShardSize uint32
```
janisz: Why do you changed visibility of this fields? This makes PR unreadable and cover main purpose of this PR.

wendigo: And breaks API.

janisz:
Closing due to lack of activity. Please reopen if you want to work on reported issues.

PR #41, opened 2017-08-04
Feature Request: support to clean expired entries

flisky:


flisky:
Emm, there is a data race indeed, but maybe it's acceptable?

flisky:
Ping, anyone?

If you don't like this, is 17e9a04 okay? I can split it out.

bigcache.go
```diff
@@ -58,6 +59,14 @@ func newBigCache(config Config, clock clock) (*BigCache, error) {
 		cache.shards[i] = initNewShard(config, onRemove)
 	}
 
+	if config.CleanWindow >= time.Second {
```
janisz: I'd prefer to check if this is ≥ 0
We need to document this behavior

config.go
```diff
@@ -8,6 +8,8 @@ type Config struct {
 	Shards int
 	// Time after which entry can be evicted
 	LifeWindow time.Duration
+	// Time tick to clean up expired entries
```
janisz: Please describe more this feature and how to disable it.

janisz:
@flisky I'm sorry for the delay. I've left some comments.

bigcache.go
flisky: Our timestamp is in unix epoch. Any tick <1 second is sort of wasting CPU.

Anyway, it's not intuitive. I'll change it.

config.go
flisky: Non-native English speaker. Any suggestion is appreciated.

bigcache.go
janisz: > Our timestamp is in unix epoch. Any tick <1 second is sort of wasting CPU.

Good point. Please add this in godoc for CleanWindow field

config.go
janisz: How about:

> // Interval between removing expired entries (clean up). If set to ≤ 0 then no action is performed. Setting to < 1 second is counterproductive — bigcache has a one second resolution.

flisky: Updated, thanks!

janisz:
@flisky Thanks!

PR #44, opened 2017-09-07
Use RLock/RUnlock for read shard.Len

cristaloleg:
I've not found any reasons why this should stay with Lock/Unlock.


janisz:
@cristaloleg Nice patch. Thanks!

PR #46, opened 2017-09-16
use defer

Leviathan1995:


janisz:
@Leviathan1995 thanks for the patch. I'd love to merge it if it does not degrade performance.

Leviathan1995:
@janisz After I benchmarking with `defer`, I think #37 is right that `defer`  operation is much slower, so this PR is meaningless. thank you!

Issue #38, opened 2017-06-27
1.9 Concurrent Maps


mxplusb:
Was looking at tip for [1.9](https://tip.golang.org), there is a new feature called [Concurrent Maps](https://tip.golang.org/doc/go1.9#sync-map), wanted to start a discussion around adding support for concurrent cache ops.

Not sure on the performance of it, but the design seems fairly solid. Looks like it uses a lot of `unsafe.Pointer` types and `atomic` ops, so it might be fairly fast.

janisz:
Looks interesting. We give it a shot and try to replace current maps and locks with concurrent map. Still we need to synchronize writes to array but reads could be faster.

mxplusb:
Yeah, I think reads could be faster, but I think the writes will be about the same, maybe ±5% since the current implementation is so highly optimized. My thoughts on this would be to try replace the base `map[uint64]uint32` implementation and see what happens. I think that and removing the explicit locking would be the only change for testing this.

If it's close in performance (±10%), it might also be useful as a secondary API instead. Something like `concurrentCache := bigcache.NewBigCache(bigcache.Config{Concurrent: true})`, where `concurrentCache` utilises the concurrent map feature, but keeps the interface the same. That being said, it would create a dependency on 1.9+, so that use case would have to be addressed.

My time for the next month is fairly limited since I'm moving, and 1.9 doesn't debut until August. I'll try and fiddle with it over the next month and see what I can come up with.

janisz:
#49 makes possible to do this task pretty easy. Just copy implementation of [`shard.go`](https://github.com/allegro/bigcache/blob/2d020b27d7c34948fdafaca969394e50959cd236/shard.go) to new file (e.g., `shard_concurent_map.go`) and replace maps and locks with concurrent maps. Finally add build tags on both files: `// +build !go1.9` and `// +build go1.9`. This will make use of concurrent hash map default in go 1.9. If the performance does not increase we can just add optional build tag to enable concurrent hash map instead of regular one.

mxplusb:
Perfect. I will work on it in the next week or so.

mxplusb:
Okay, this is actually going to be a big change as concurrent maps do not support indexing. So this will take awhile.

PR #47, opened 2017-09-19
Extract Lock/Unlock from shard

cristaloleg:
Looking at all usages of a shard, locks are assessed outside, but `reset` method still uses lock inside.
There is no gain in terms of performance, just consistency of the code.

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -134,7 +134,9 @@ func (c *BigCache) Set(key string, entry []byte) error {
 // Reset empties all cache shards
 func (c *BigCache) Reset() error {
 	for _, shard := range c.shards {
+		shard.lock.Lock()
 		shard.reset(c.config)
+		shard.lock.Unlock()
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


cristaloleg:
Coverage fail isn't related to my changes, it was broken before me :P

janisz:
LGTM @druminski What do you think? W do not lock in any other shard function it's good so this PR really make our code more consistent.

cristaloleg:
To be honest it will be better to move all operations with locks to the shards, this will reduce dependence on their(shards) implementation. This will allow us to use different shards(sync.Map or atomic map or smth else). If team will agree on this I can work on this.

janisz:
Great idea! We can make it step by step. First, create an interface for shard so we get prevented from messing with its internal state as we are doing now. Basically, we need to hide `entries` and provide functions to work with them. Implementation should be threadsafe. Finally, we can provide different implementations, e.g., as you write based on `sync.Map` proposed in #38.

cristaloleg:
Yep, totally agree, that’s what I would like to implement. Can I start? :)

janisz:
Cool. I'll be happy to merge it.

druminski:
@cristaloleg @janisz LGTM, you can also try to extract shard interface, have fun :)

PR #48, opened 2017-09-19
Get rid of defer in interator

cristaloleg:
One more PR for code consistency. This one also adds small performance gain:
```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkIterateOverCache/512-shards-8      495           423           -14.55%
BenchmarkIterateOverCache/1024-shards-8     487           424           -12.94%
BenchmarkIterateOverCache/8192-shards-8     470           389           -17.23%
```


janisz:
@cristaloleg Thanks!

PR #49, opened 2017-09-21
Improve encapsulation of shards

cristaloleg:
- `Get`, `Set` methods are moved from `bigcache.go` to `shard.go`
- `initNewShard` has additional param `clock`
- all locks logic exists only inside shard
- get rid of `copyCurrentShardMap`
- `getIndex` and `oldest` methods added to shard
- introduced index variable in `getShard`

Next PR will introduce `Shard` interface, `newShard` method and additional changes.

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -2,7 +2,6 @@ package bigcache
 
 import (
 	"fmt"
-	"log"
 	"time"
 )
 
@@ -56,7 +55,7 @@ func newBigCache(config Config, clock clock) (*BigCache, error) {
 	}
 
 	for i := 0; i < config.Shards; i++ {
-		cache.shards[i] = initNewShard(config, onRemove)
+		cache.shards[i] = initNewShard(config, onRemove, clock)
 	}
 
 	if config.CleanWindow > 0 {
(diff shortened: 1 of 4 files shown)


cristaloleg:
As before: coverage failure was introduced before.

shard.go
```diff
@@ -12,10 +14,112 @@ type cacheShard struct {
...
+
+func (s *cacheShard) getIndex(index int) ([]byte, error) {
+	return s.entries.Get(index)
+}
+
+func (s *cacheShard) copyKeys() ([]uint32, int) {
+	s.lock.RLock()
+
+	var elements = make([]uint32, len(s.hashmap))
+	next := 0
```
janisz: `next` is local variable and could be defined outside of critical section (before `s.lock.RLock()`)

shard.go
```diff
@@ -12,10 +14,112 @@ type cacheShard struct {
...
+
+func (s *cacheShard) oldest() ([]byte, error) {
+	return s.entries.Peek()
+}
+
+func (s *cacheShard) getIndex(index int) ([]byte, error) {
+	return s.entries.Get(index)
+}
+
+func (s *cacheShard) copyKeys() ([]uint32, int) {
```
janisz: Return values of this function are not obvious (espsecialy second → `next`) can you add a godoc with information what is the result of this function.

janisz: It looks like it wasn't documented before. How about using named return parameters to defined what's returned?

```
func (s *cacheShard) copyKeys() (keyes []uint32, length int) {
```

bigcache.go
```diff
@@ -171,20 +116,13 @@ func (c *BigCache) onEvict(oldestEntry []byte, currentTimestamp uint64, evict fu
...
-		}
-		shard.lock.Unlock()
+		shard.cleanUp(currentTimestamp)
 	}
 }
 
 func (c *BigCache) getShard(hashedKey uint64) (shard *cacheShard) {
-	return c.shards[hashedKey&c.shardMask]
+	index := hashedKey & c.shardMask
+	return c.shards[index]
```
janisz: This could be left untouched.

shard.go
```diff
@@ -12,10 +14,112 @@ type cacheShard struct {
...
+		}
+	}
+	s.lock.Unlock()
+}
+
+func (s *cacheShard) oldest() ([]byte, error) {
+	return s.entries.Peek()
+}
+
+func (s *cacheShard) getIndex(index int) ([]byte, error) {
```
janisz: I think this should be renamed. When I see `getIndex` I think that this function will return an index. How about `getEntryUnderIndex` or just `getEntry`?

shard.go
```diff
@@ -12,10 +14,112 @@ type cacheShard struct {
...
+		if oldestEntry, err := s.entries.Peek(); err != nil {
+			break
+		} else if evicted := s.onEvict(oldestEntry, currentTimestamp, s.removeOldestEntry); !evicted {
+			break
+		}
+	}
+	s.lock.Unlock()
+}
+
+func (s *cacheShard) oldest() ([]byte, error) {
```
janisz: How about renaming to `getOldestEntry`? Below is a method with `get...` so we should follow one naming scheme and best if functions are named as a verb declaring what they are doing. I'm fine with a convention of naming function with a noun that they produce but then we need to rename all other functions.

cristaloleg:
Thanks, will update tomorrow.
WDYT about creating a release on Github before big changes will be merged?

janisz:
Thanks for quick response. We haven't created any release so far. It's a good idea to start making releases changelog for users. IMO API is stable and we definitely want to keep backward compatibility so we can start with version 1.0. 

@druminski What do you think?

druminski:
@janisz Yes, I do agree with you that the API is stable, beside that we don't have plans to do major changes. Therefore we can start release version 1.0. As for writing a changelog, imo it is a good idea.

@cristaloleg overall your PR looks good, although please apply @janisz suggestions

cristaloleg:
@janisz  @druminski I'll update PR later today, thank for the review.

janisz:
@cristaloleg LGTM. I'll create a release and merge it after.

cristaloleg:
@janisz ok, as you wish.

janisz:
Done. @cristaloleg  Huge thanks for this refactoring.

cristaloleg:
@janisz no problem, will push next PR in a few days.

PR #50, opened 2017-10-08
Adding logging interface & default logger

### LINKED ISSUE #36: Custom logging
It would be nice if you could pass a custom logger when setting `bigcache.Config.Verbose`.

lxfontes:
👋 ohai !

This maintains backwards compatibility in terms of logging format, also keeps `Verbose` as main toggle.

IMO: it would look better to 🔪  `Verbose` , keep only `Logger` and have a default noop-logger. However, that means breaking api compatibility.

Fixes #36

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -10,6 +10,9 @@ type Logger interface {
 	Printf(format string, v ...interface{})
 }
 
+// this is a safeguard, breaking on compile time in case
+// `log.Logger` does not adhere to our `Logger` interface.
+// see https://golang.org/doc/faq#guarantee_satisfies_interface
 var _ Logger = &log.Logger{}
 
 // DefaultLogger returns a `Logger` implementation
--- a/logger.go
+++ b/logger.go
@@ -1,0 +1,27 @@
+package bigcache
+
+import (
+	"log"
+	"os"
+)
+
+// Logger is invoked when `Config.Verbose=true`
+type Logger interface {
+	Printf(format string, v ...interface{})
+}
+
+var _ Logger = &log.Logger{}
+
+// DefaultLogger returns a `Logger` implementation
+// backed by stdlib's log
+func DefaultLogger() *log.Logger {
+	return log.New(os.Stdout, "", log.LstdFlags)
+}
+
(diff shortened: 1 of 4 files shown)


logger.go
```diff
@@ -0,0 +1,27 @@
...
+// Logger is invoked when `Config.Verbose=true`
+type Logger interface {
+	Printf(format string, v ...interface{})
+}
+
+var _ Logger = &log.Logger{}
+
+// DefaultLogger returns a `Logger` implementation
+// backed by stdlib's log
+func DefaultLogger() *log.Logger {
```
janisz: I'm not sure if this should be a function or a package variable. Just like [net.http.DefaultClient](https://golang.org/pkg/net/http/#pkg-variables)

logger.go
```diff
@@ -0,0 +1,27 @@
...
+	"log"
+	"os"
+)
+
+// Logger is invoked when `Config.Verbose=true`
+type Logger interface {
+	Printf(format string, v ...interface{})
+}
+
+var _ Logger = &log.Logger{}
```
janisz: Why do we need it?

janisz:
@lxfontes LGTM. 

I agree we should cut `Verbose` but we should do it in a way to keep backward compatibility e.g., If verbose is false then set logger to noop logger. We can handle it in next pull request.

logger.go
lxfontes: This will emit a compile error in case stdlib's `log.Logger` fails to implement our logger interface.

https://golang.org/doc/faq#guarantee_satisfies_interface

logger.go
lxfontes: tots agree 👍 
Idea was to not rely global state. global vars allow things like:

```
package main

import "net/http"

func main() {
	http.DefaultClient = &http.Client{}
}
```

hence the function route

logger.go
janisz: OK, that's what I suspected. Can you add a comment on this with a link. Just in case somebody wants to delete this line in a future.

@medzin What do you think?

janisz:
@lxfontes Thanks :clap:

PR #53, opened 2017-10-23
implement bigcache.Delete

### LINKED ISSUE #35: No way to increment value without updating the timestamp
These commonly used methods are missing:

1) Increment/Decrement integer-based value without forcing a set, which puts a new timestamp - this is a problem in that, when I need to collect counts for keys which are popular, they may never expire (lets say expiration/eviction triggers a callback function which writes count value every "5 minutes" to log file)

2) Delete key - there is not way to remove a key or force eviction function on particular keys, overwriting assumes one has a new value

Nice to have: When calling "Reset" would be nice to trigger OnRemove callback or to provide another function which does that

flisky:
I'm trying to update the individual elements on necessary, and `emulating with spacial value` is not a perfect choice to me.

Fix #35 partially.

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -83,6 +83,13 @@ func (c *BigCache) Set(key string, entry []byte) error {
 	return shard.set(key, hashedKey, entry)
 }
 
+// Del removes the key
+func (c *BigCache) Del(key string) error {
+	hashedKey := c.hash.Sum64(key)
+	shard := c.getShard(hashedKey)
+	return shard.del(key, hashedKey)
+}
+
 // Reset empties all cache shards
 func (c *BigCache) Reset() error {
 	for _, shard := range c.shards {
--- a/bigcache.go
+++ b/bigcache.go
@@ -83,8 +83,8 @@ func (c *BigCache) Set(key string, entry []byte) error {
 	return shard.set(key, hashedKey, entry)
 }
 
-// Del removes the key
-func (c *BigCache) Del(key string) error {
+// Delete removes the key
+func (c *BigCache) Delete(key string) error {
 	hashedKey := c.hash.Sum64(key)
 	shard := c.getShard(hashedKey)
 	return shard.del(key, hashedKey)
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -194,14 +194,14 @@ func TestCacheDel(t *testing.T) {
 	cache, _ := NewBigCache(DefaultConfig(time.Second))
 
 	// when
-	err := cache.Del("nonExistingKey")
+	err := cache.Delete("nonExistingKey")
 
 	// then
 	assert.Equal(t, err.Error(), "Entry \"nonExistingKey\" not found")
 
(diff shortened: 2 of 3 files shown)


flisky:
Ping @druminski.

bigcache_test.go
```diff
@@ -187,6 +187,26 @@ func TestCacheLen(t *testing.T) {
...
+
+	// then
+	assert.Equal(t, err.Error(), "Entry \"nonExistingKey\" not found")
+
+	// and when
+	cache.Set("existingKey", nil)
+	err = cache.Del("existingKey")
+
+	// then
+	assert.Nil(t, err)
```
janisz: Can you assert if the key is actually deleted?

shard.go
```diff
@@ -76,6 +76,28 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
...
+
+	if itemIndex == 0 {
+		s.lock.RUnlock()
+		return notFound(key)
+	}
+
+	wrappedEntry, err := s.entries.Get(int(itemIndex))
+	if err != nil {
+		s.lock.RUnlock()
+		return err
```
janisz: Is there a test for this error?

bigcache_test.go
janisz: Can you assert if the key is actually deleted? Key should not exist and `Len() == 0`.

shard.go
janisz: Please, add a test for this case.

janisz: Please, add a test for this case.

bigcache_test.go
janisz: Can you assert if the key is actually deleted? Key should not exist and `Len() == 0`.

shard.go
flisky: Sorry, it's really hard to test this case.

There're three error cases for `queue.BytesQueue#Get`:

1. `itemIndex <= 0`: 0 already covered and `uint` never negative.
2. empty queue & 3 index size overflow have no way to be faked.

Actually, the [same logic](https://coveralls.io/builds/13894823/source?filename=shard.go#L36) in `bigcache#Get` is also not unittest covered.

Any suggestions?

bigcache.go
```diff
@@ -83,6 +83,13 @@ func (c *BigCache) Set(key string, entry []byte) error {
 	return shard.set(key, hashedKey, entry)
 }
 
+// Del removes the key
+func (c *BigCache) Del(key string) error {
```
druminski: Can we change "Del" to "Delete"? Shortcut can be the source of misunderstanding or doubts about what function does when we see only the name. Therefore I prefer full name over shortcut, especially when the function is exposed in the API. Beside that I don't have any suggestions, the code looks great, thanks for this PR.

flisky:
Ping @janisz

janisz:
I'm sorry for the delay. I was out of town. @filsky Thanks for contribution!

PR #57, opened 2017-11-14
Cache stats

cristaloleg:


shard.go
```diff
@@ -69,9 +76,10 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
 			s.hashmap[hashedKey] = uint32(index)
 			s.lock.Unlock()
 			return nil
-		} else if s.removeOldestEntry() != nil {
+		}
+		if s.removeOldestEntry() != nil {
 			s.lock.Unlock()
-			return fmt.Errorf("Entry is bigger than max shard size.")
+			return fmt.Errorf("entry is bigger than max shard size")
```
cristaloleg: Fixed annoying linter warning.

shard.go
```diff
@@ -167,6 +178,10 @@ func (s *cacheShard) len() int {
 	return res
 }
 
+func (s *cacheShard) getStats() Stats {
+	return s.stats
```
cristaloleg: Should we use `atomic.LoadInt64` here instead of raw copy?

shard.go
```diff
@@ -28,22 +31,26 @@ func (s *cacheShard) get(key string, hashedKey uint64) ([]byte, error) {
 
 	if itemIndex == 0 {
 		s.lock.RUnlock()
+		atomic.AddInt64(&s.stats.Misses, 1)
```
janisz: How about internal method e.g., `s.stats.miss()`

bigcache.go
```diff
@@ -107,6 +107,19 @@ func (c *BigCache) Len() int {
 	return len
 }
 
+// Stats returns cache's statistics
+func (c *BigCache) Stats() Stats {
```
janisz: Can you add a test for this method? Best if you can extend current tests with additional assertion

cristaloleg:
@janisz I did benchmark and the biggest difference was 30ns for `BenchmarkWriteToUnlimitedCacheWithSmallInitSizeAnd1Shard-8`
No new allocs and few bytes in memory usage(not in all benchmarks)

cristaloleg:
Merging or smth left to fix?

shard.go
```diff
@@ -28,22 +31,26 @@ func (s *cacheShard) get(key string, hashedKey uint64) ([]byte, error) {
...
 		s.lock.RUnlock()
+		s.miss()
 		return nil, err
 	}
 	if entryKey := readKeyFromEntry(wrappedEntry); key != entryKey {
 		if s.isVerbose {
 			s.logger.Printf("Collision detected. Both %q and %q have the same hash %x", key, entryKey, hashedKey)
 		}
 		s.lock.RUnlock()
+		s.miss()
```
mxplusb: I'm not sure this is a miss as much as it's a legitimate error. Maybe skip this one?

cristaloleg: Make sense, would like to hear opinions from other contributors.

janisz: How about adding a dedicated metric for collisions? This might be useful for experiments with different hashing algorithms.

janisz:
@cristaloleg Thanks!

PR #58, opened 2017-11-16
Added basic HTTP server implementation.

mxplusb:
Features:
Basic HTTP server.
Basic middleware implementation.

I wanted to provide some thoughts on this implementation.

API Surface:

```
GET /api/v1/cache/{key}
PUT /api/v1/cache/{key}
```

The API path denotes it's an API which is versioned, so making changes/implementing new features allows it to be versioned easily. I felt `PUT` was more relevant than `POST` because `bigcache.BigCache.Set()` will overwrite an existing key of the same name. `PUT` more closely aligns with the semantic aspect of the feature as opposed to the spirit. When looking at the `/cache/` heirarchy of the path, there was a PR I saw recently which looked to implement stats about the cache. If it gets approved, you can add `/api/v1/stats` to the API surface as an extra API. Overall, this API surface seemed easily sustainable to me.

HTTP Implementation:

What I love about this package is the focus is on performance and the standard library. My proposal suggested using the standard library, so I felt that was important. When looking at implementation options, I wanted something that could be easily extensible without pulling in dependencies. This [post](https://medium.com/@matryer/writing-middleware-in-golang-and-how-go-makes-it-so-much-fun-4375c1246e81) gave me a great idea, and I liked the adapter idea because it's highly extensible using only the standard library. If you want to implement request tracing, it's easy to do by adding a quick service adapter; if you want to implement some form of authentication or geo-limitation, just add another service adapter. The `requestMetrics()` feature, to me, is both a way to log basic metrics about the HTTP server, as well as an example for a basic service adapter.

Overall, my goal with this implementation was a focus on the standard library with no external dependencies in a highly extensible way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .idea
 .DS_Store
+/server/server.exe
+/server/server
--- a/server/README.md
+++ b/server/README.md
@@ -1,0 +1,91 @@
+# BigCache HTTP Server
+
+This is a basic HTTP server implementation for BigCache. It has a basic RESTful API.
+
+```bash
+GET /api/v1/cache/{key}
+PUT /api/v1/cache/{key}
+```
+
+This server is designed to be consumed as a standalone executable, for things like Cloud Foundry, Heroku, etc.
+
+### Command-line Interface
+
+```powershell
+PS C:\go\src\github.com\mxplusb\bigcache\server> .\server.exe -h
+Usage of C:\go\src\github.com\mxplusb\bigcache\server\server.exe:
+  -lifetime int
+        Lifetime, in minutes, of each cache object. (default 10)
+  -logfile string
+        Location of the logfile.
+  -max int
+        Maximum amount of objects in the cache. (default 8192)
+  -maxInWindow int
+        Used only in initial memory allocation. (default 600000)
+  -maxShardEntrySize int
+        The maximum size of each object stored in a shard. Used only in initial memory allocation. (default 500)
+  -port int
+        The port to listen on. (default 9090)
+  -shards int
+        Number of shards for the cache. (default 1024)
+  -v    Verbose logging.
+
+```
+
+Example:
+
+```bash
+$ curl -v -XPUT localhost:9090/api/v1/cache/example -d "yay!"
+*   Trying 127.0.0.1...
+* Connected to localhost (127.0.0.1) port 9090 (#0)
+> PUT /api/v1/cache/example HTTP/1.1
+> Host: localhost:9090
+> User-Agent: curl/7.47.0
+> Accept: */*
+> Content-Length: 4
+> Content-Type: application/x-www-form-urlencoded
+>
+* upload completely sent off: 4 out of 4 bytes
+< HTTP/1.1 201 Created
+< Date: Fri, 17 Nov 2017 03:50:07 GMT
+< Content-Length: 0
+< Content-Type: text/plain; charset=utf-8
+<
+* Connection #0 to host localhost left intact
+$ 
+$ curl -v -XGET localhost:9090/api/v1/cache/example
+Note: Unnecessary use of -X or --request, GET is already inferred.
+*   Trying 127.0.0.1...
+* Connected to localhost (127.0.0.1) port 9090 (#0)
+> GET /api/v1/cache/example HTTP/1.1
+> Host: localhost:9090
+> User-Agent: curl/7.47.0
+> Accept: */*
+>
+< HTTP/1.1 200 OK
+< Date: Fri, 17 Nov 2017 03:50:23 GMT
+< Content-Length: 4
+< Content-Type: text/plain; charset=utf-8
+<
+* Connection #0 to host localhost left intact
+yay!
+```
+
+The server does log basic metrics:
+
+```bash
+$ ./server
+2017/11/16 22:49:22 cache initialised.
+2017/11/16 22:49:22 starting server on :9090
+2017/11/16 22:50:07 stored "example" in cache.
+2017/11/16 22:50:07 request took 277000ns.
+2017/11/16 22:50:23 request took 9000ns.
+```
+
+### Acquiring
+
+This is native Go with no external dependencies, so it will compile for all supported Golang platforms. To build:
+
+```bash
+go build server.go
+```
--- a/server/server.go
+++ b/server/server.go
@@ -1,0 +1,157 @@
+package main
+
+import (
+	"flag"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/allegro/bigcache"
+)
+
+const (
+	// base HTTP paths.
+	apiVersion  = "v1"
+	apiBasePath = "/api/" + apiVersion + "/"
+
+	// path to cache.
+	cachePath = apiBasePath + "cache/"
+)
+
+var (
+	port    int
+	logfile string
+
+	// cache-specific settings.
+	cache  *bigcache.BigCache
+	config = bigcache.Config{}
+)
+
+func init() {
+	flag.BoolVar(&config.Verbose, "v", false, "Verbose logging.")
+	flag.IntVar(&config.Shards, "shards", 1024, "Number of shards for the cache.")
+	flag.IntVar(&config.MaxEntriesInWindow, "maxInWindow", 1000*10*60, "Used only in initial memory allocation.")
+	flag.DurationVar(&config.LifeWindow, "lifetime", 100000*100000*60, "Lifetime of each cache object.")
+	flag.IntVar(&config.HardMaxCacheSize, "max", 8192, "Maximum amount of data in the cache in MB.")
+	flag.IntVar(&config.MaxEntrySize, "maxShardEntrySize", 500, "The maximum size of each object stored in a shard. Used only in initial memory allocation.")
+	flag.IntVar(&port, "port", 9090, "The port to listen on.")
+	flag.StringVar(&logfile, "logfile", "", "Location of the logfile.")
+}
+
+func main() {
+	flag.Parse()
+
+	var logger *log.Logger
+
+	if logfile == "" {
+		logger = log.New(os.Stdout, "", log.LstdFlags)
+	} else {
+		f, err := os.OpenFile(logfile, os.O_APPEND|os.O_WRONLY, 0600)
+		if err != nil {
+			panic(err)
+		}
+		logger = log.New(f, "", log.LstdFlags)
+	}
+
+	var err error
+	cache, err = bigcache.NewBigCache(config)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	logger.Print("cache initialised.")
+
+	// let the middleware log.
+	http.Handle(cachePath, serviceLoader(cacheIndexHandler(), requestMetrics(logger)))
+
+	logger.Printf("starting server on :%d", port)
+
+	strPort := ":" + strconv.Itoa(port)
+	log.Fatal("ListenAndServe: ", http.ListenAndServe(strPort, nil))
+}
+
+// our base middleware implementation.
+type service func(http.Handler) http.Handler
+
+// chain load middleware services.
+func serviceLoader(h http.Handler, svcs ...service) http.Handler {
+	for _, svc := range svcs {
+		h = svc(h)
+	}
+	return h
+}
+
+func cacheIndexHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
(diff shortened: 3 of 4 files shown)


server/server.go
```diff
@@ -0,0 +1,157 @@
...
+// handles get requests.
+func getCacheHandler(w http.ResponseWriter, r *http.Request) {
+	target := r.URL.Path[len(cachePath):]
+	if target == "" {
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte("can't get a key if there is no key."))
+		log.Print("empty request.")
+		return
+	}
+	entry, err := cache.Get(target)
```
mxplusb: I tried to do a `string.Contains(err.Error(), "not found")` search on the error so I could give a 404 instead of a 500 if an item didn't exist, but I kept running into nil reference errors when trying to read the error message. If you know a way to do that, I can totally implement it as part of this.

server/server.go
```diff
@@ -0,0 +1,157 @@
...
+			panic(err)
+		}
+		logger = log.New(f, "", log.LstdFlags)
+	}
+
+	config := bigcache.Config{
+		Shards:             1024,
+		LifeWindow:         time.Duration(lifeWindow) * time.Minute,
+		MaxEntriesInWindow: 1000 * 10 * 60,
+		MaxEntrySize:       500,
```
janisz: How about making `Shards`, `MaxEntriesInWindow` and `MaxEntriesInWindow` configurable too?

server/server.go
janisz: You may first check if `err` is not `nil` and if not then use `string.Contains`

server/server.go
```diff
@@ -0,0 +1,157 @@
...
+		return
+	}
+
+	if err := cache.Set(target, []byte(entry)); err != nil {
+		log.Print(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	log.Printf("stored \"%s\" in cache.", target)
+	w.WriteHeader(http.StatusAccepted)
```
janisz: I think this should return `200` value is stored. It's a blocking call.

mxplusb: I read the [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) and yeah, 202 is not the right code, I can fix that. I think it should actually be 201 instead of 200 since we are creating a cache entry. Thoughts?

server/server.go
mxplusb: Sure, I don't see why not.

mxplusb:
I can add tests and a README. Gimme a few days.

server/server.go
janisz: :+1: 201 looks good

mxplusb:
Here's what I get when I try and check if the error message:

```
--- FAIL: TestGetWithMissingKey (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x30 pc=0x627abd]

goroutine 11 [running]:
testing.tRunner.func1(0xc0420fc1e0)
        C:/Go/src/testing/testing.go:711 +0x2d9
panic(0x66e260, 0x819200)
        C:/Go/src/runtime/panic.go:491 +0x291
github.com/allegro/bigcache.(*BigCache).Get(0x0, 0xc0420123a5, 0xc, 0x8240e0, 0x3130d50, 0x0, 0x6a47a0, 0xc06716ab70)
        C:/Users/MikeLloyd/Development/go/src/github.com/allegro/bigcache/bigcache.go:74 +0x2d
_/C_/Users/MikeLloyd/Development/go/src/github.com/mxplusb/bigcache/server.getCacheHandler(0x7ef5a0, 0xc04204ea00, 0xc0682fe
000)
        C:/Users/MikeLloyd/Development/go/src/github.com/mxplusb/bigcache/server/server.go:132 +0x8d
_/C_/Users/MikeLloyd/Development/go/src/github.com/mxplusb/bigcache/server.TestGetWithMissingKey(0xc0420fc1e0)
        C:/Users/MikeLloyd/Development/go/src/github.com/mxplusb/bigcache/server/server_test.go:47 +0x186
testing.tRunner(0xc0420fc1e0, 0x6cf768)
        C:/Go/src/testing/testing.go:746 +0xd7
created by testing.(*T).Run
        C:/Go/src/testing/testing.go:789 +0x2e5
exit status 2
```

For a reason I haven't figured out yet, if I dereference the error, like this:

```
entry, err := cache.Get(target)
	if err != nil {
		errMsg := (err).Error()
		if strings.Contains(errMsg, "not found") {
			log.Print(err)
			w.WriteHeader(http.StatusNotFound)
			return
		}
		log.Print(err)
		w.WriteHeader(http.StatusInternalServerError)
		return
	}
```

It works. If anyone can explain why, I'd be interesting in learning. I know it's because `cache` is a pointer, but I don't understand the need to dereference the error.

server/server.go
mxplusb: Implemented.

mxplusb:
Alright, it's been updated with tests and a basic README.

server/server.go
mxplusb: Turns out I had to dereference the error.

server/server_test.go
```diff
@@ -0,0 +1,96 @@
...
+		LifeWindow:         10 * time.Minute,
+		MaxEntriesInWindow: 1000 * 10 * 60,
+		MaxEntrySize:       500,
+		Verbose:            true,
+		HardMaxCacheSize:   8192,
+		OnRemove:           nil,
+	})
+}
+
+func TestGetWithNoKey(t *testing.T) {
```
janisz: It will be great if we can run tests in parallel to ensure we do not rely on their order and they do not interfere with each other  https://golang.org/pkg/testing/#T.Parallel

server/server.go
```diff
@@ -0,0 +1,170 @@
...
+var (
+	shards             int
+	maxEntriesInWindow int
+	lifeWindow         int64
+	hardMax            int
+	verbose            bool
+	maxEntrySize       int
+	logfile            string
+	port               int
+	cache              *bigcache.BigCache
```
janisz: I prefer to not have a global cache. It makes testing harder

server/server.go
```diff
@@ -0,0 +1,170 @@
...
+
+var (
+	shards             int
+	maxEntriesInWindow int
+	lifeWindow         int64
+	hardMax            int
+	verbose            bool
+	maxEntrySize       int
+	logfile            string
+	port               int
```
janisz: Instead of declaring fields for every config entry we can just use config struct see https://github.com/allegro/marathon-consul/blob/1.4.2/config/config.go#L63

server/server.go
```diff
@@ -0,0 +1,170 @@
...
+	config := bigcache.Config{
+		Shards:             shards,
+		LifeWindow:         time.Duration(lifeWindow) * time.Minute,
+		MaxEntriesInWindow: maxEntriesInWindow,
+		MaxEntrySize:       maxEntrySize,
+		Verbose:            verbose,
+		HardMaxCacheSize:   hardMax,
+		OnRemove:           nil,
+		Logger:             logger,
+	}
```
janisz: this could be removed with global cache config.

server/server_test.go
```diff
@@ -0,0 +1,96 @@
...
+		t.FailNow()
+	}
+
+	if string(body) != "123" {
+		t.Logf("want: 123; got: %s.\n\tcan't get existing key getKey.", string(body))
+		t.Fail()
+	}
+}
+
+func TestPutKey(t *testing.T) {
```
janisz: Missing case when there is no key in a request.

server/README.md
```diff
@@ -0,0 +1,91 @@
...
+2017/11/16 22:50:07 request took 277000ns.
+2017/11/16 22:50:23 request took 9000ns.
+```
+
+### Acquiring
+
+To build:
+
+```bash
+GOOS=windows go build server.go
```
janisz: `GOOS` is redundant in most cases

server/server.go
mxplusb: I followed the [`database/sql` pattern](https://stackoverflow.com/questions/17376207/how-to-share-mysql-connection-between-http-goroutines) for multiple HTTP handlers as access will need to be shared across multiple handlers if it's going to be extensible for stats and other future features. How would you prefer cache access?

server/server_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"bytes"
+	"io/ioutil"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/allegro/bigcache"
+)
+
+var (
```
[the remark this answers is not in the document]
mxplusb: Fixed

server/server_test.go
mxplusb: Fixed. Ran `go test -v .` in a loop 25 times to make sure.

server/server_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		t.Log(err)
+		t.FailNow()
+	}
+
+	if string(body) != "123" {
+		t.Logf("want: 123; got: %s.\n\tcan't get existing key getKey.", string(body))
+		t.Fail()
```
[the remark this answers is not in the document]
mxplusb: Fixed.

server/server_test.go
```diff
@@ -0,0 +1,96 @@
...
+	testCacheSetup()
+	req := httptest.NewRequest("PUT", testBaseString+"/api/v1/cache/putKey", bytes.NewBuffer([]byte("123")))
+	rr := httptest.NewRecorder()
+
+	putCacheHandler(rr, req)
+
+	testPutKeyResult, err := cache.Get("putKey")
+	if err != nil {
+		t.Log(err)
+		t.Fail()
```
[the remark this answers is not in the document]
mxplusb: Fixed.

server/server_test.go
mxplusb: Fixed.

server/server.go
mxplusb: Fixed.

mxplusb:
Fixed requested changes minus the global cache via amendment.

@janisz can you provide some thoughts on how to share a non-global cache between handlers? I'm not sure what you're looking for.

server/server.go
```diff
@@ -0,0 +1,160 @@
...
+
+var (
+	lifeWindow int64
+	port       int
+	logfile    string
+
+	// cache-specific settings.
+	cache  *bigcache.BigCache
+	config = bigcache.Config{
+		LifeWindow: time.Duration(lifeWindow) * time.Minute,
```
janisz: Why it's not configurable? https://golang.org/pkg/flag/#DurationVar

janisz:
@mxplusb I added one  comment.

I think we can leave global cache. I was thinking about creating the cache in main and then pass it as an argument to the function but this will only complicate code without any improvement in testability.

server/server.go
mxplusb: I didn't realise there was a time.Duration option in that package, to be honest. I will update with that.

mxplusb:
Fixed last configuration review request with amendment.

.gitignore
```diff
@@ -1,2 +1,3 @@
 .idea
-.DS_Store
\ No newline at end of file
+.DS_Store
+/server/server.exe
```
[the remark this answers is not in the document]
janisz: :+1:

janisz:
@mxplusb Would you like to fix small issues reported by @cristaloleg ?

cristaloleg:
They aren't so important 😉

mxplusb:
Sorry, last week was a holiday week in the US (Thanksgiving), so I was focused on that. I will take a look at the tasks this week and get them addressed.

.gitignore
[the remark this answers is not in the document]
mxplusb: I will add this.

server/server.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/allegro/bigcache"
+)
+
+const (
+	// base HTTP paths.
+	apiVersion  = "v1"
+	apiBasePath = "/api/" + apiVersion + "/"
+
+	// path to cache.
```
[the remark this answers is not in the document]
mxplusb: My next PR will contain the rest of the cache API surface area (finished recently), so there is a `statsPath` constant as well for the cache stats, with a similar comment. Personally I'd leave it for others, but it really doesn't matter to me. I'll remove it.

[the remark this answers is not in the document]
cristaloleg: you can leave it ;)

mxplusb:
Fixed requested changes. 

//cc @janisz @cristaloleg

janisz:
@mxplusb Thanks for your contribution.
@cristaloleg Thank you for your time.

PR #68, opened 2018-03-18
Fix some race conditions

cristaloleg:
1. Run every package on travis
2. No more race condition in `shard.GetStats`
3. No more race condition in `shard.del`
4. New `TestParallel` test to have `Set`, `Get`, `Del` operations running in parallel.


cristaloleg:
I've a proposal to disable coverage for `server` package :D

PR #69, opened 2018-03-21
fixed gc bug and updated readme.

### LINKED ISSUE #59: GC Pause time for maps.
I understand that benchmarks will depend on the machine used.
Following is from a 2017 Macbook pro (2.8 GHz Intel Core i7, 16 GB 2133 MHz LPDDR3)

However native Maps seems to have had a massive improvement. Is this due to go 1.9?

`go version go1.9.2 darwin/amd64`

```
go run caches_gc_overhead_comparison.go
Number of entries:  20000000
GC pause for bigcache:  45.222µs
GC pause for freecache:  37.043µs
GC pause for map:  59.572µs
```

mxplusb:
Looking at #59, I wanted to do a quick once over of the comparisons in
1.10 so we can make sure our numbers are still accurate. When looking
through the GC code, I discovered we are using the wrong GC stat. Using
debug.GCStats.Pause[0] uses the most recent GC pause, which can be zero
(as I found out on Windows). We should be using total pause for all
collections, which is debug.GCStats.PauseTotal. That will show the total
length of time for all GC activities during our test, which more
accurately reflects the type of comparison we are trying to show.

Also, I updated the readme to reflect our GC comparison with go1.10.

fixes #59

### DIFF
--- a/caches_bench/caches_gc_overhead_comparison.go
+++ b/caches_bench/caches_gc_overhead_comparison.go
@@ -14,7 +14,7 @@ func gcPause() time.Duration {
 	runtime.GC()
 	var stats debug.GCStats
 	debug.ReadGCStats(&stats)
-	return stats.Pause[0]
+	return stats.PauseTotal
 }
 
 const (
--- a/README.md
+++ b/README.md
@@ -99,9 +99,9 @@ Writes to map are the slowest.
 cd caches_bench; go run caches_gc_overhead_comparison.go
 
 Number of entries:  20000000
-GC pause for bigcache:  27.81671ms
-GC pause for freecache:  30.218371ms
-GC pause for map:  11.590772251s
+GC pause for bigcache:  5.8658ms
+GC pause for freecache:  32.4341ms
+GC pause for map:  52.9661ms
 ```
 
 Test shows how long are the GC pauses for caches filled with 20mln of entries.


janisz:
@mxplusb Thanks!

buraksezer:
"It is clear that both reduce GC overhead in contrast to map which GC pause time took more than **10 seconds.**"

That's not right anymore. It just takes 52.9ms in your test. Not 10 seconds.

caches_bench/caches_gc_overhead_comparison.go
```diff
@@ -14,7 +14,7 @@ func gcPause() time.Duration {
 	runtime.GC()
 	var stats debug.GCStats
 	debug.ReadGCStats(&stats)
-	return stats.Pause[0]
+	return stats.PauseTotal
```
janisz: This change makes our all benchmarks wrong :(
Using the latest GC pause was done on purpose. Here we get gc pause for the whole application so in current code, GC pause for Map will be always higher than GC pause for BigCache since it will sum real map gc pause and bigcache gc pause. To prevent this we should subtract previously observed GC Pause.

janisz: https://github.com/allegro/bigcache-bench/issues/7

PR #72, opened 2018-03-23
updated .gitignore

mxplusb:
once this is pushed I will push the tag and then create the new release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 /server/server.exe
 /server/server
+CHANGELOG.md


.gitignore
```diff
@@ -2,3 +2,4 @@
 .DS_Store
 /server/server.exe
 /server/server
+CHANGELOG.md
```
janisz: Why do we need to ignore CHANGELOG.md? There is no such file and if we add it should not be ignored

mxplusb: I use [this tool](https://github.com/skywinder/github-changelog-generator) to generate the release notes. It creates/updates a CHANGELOG.md file every time it's run, and I just want to ignore it so I don't have to delete it/think about it if I run it more than once.

PR #74, opened 2018-03-26
Add tests to prove/enforce 0-value byte slice caching

jshufro:
My team is currently using this library for a new application- essentially just a cache that sits in front of one of our slower, on-disk data stores and holds a portion of the data for faster access.

One of our requirements is that we can cache failed reads from the database- ie, a read where the key is not found in the database. This is intended to reduce traffic to the database when a large portion of the keys are not found.

I implemented his by caching the key with a nil value- something bigcache implicitly supports already. I figured it would be a decent idea to make the support more explicit with some tests.

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -511,6 +511,43 @@ func TestHashCollision(t *testing.T) {
 	assert.Equal(t, cache.Stats().Collisions, int64(1))
 }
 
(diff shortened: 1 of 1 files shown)


mxplusb:
Travis looks happy, tests seem good to me, I'm good with it.

jshufro:
No problem :) Thanks for the speedy merges and great library.

PR #76, opened 2018-03-28
added docker reference.

mxplusb:
Added initial docker image reference. does not use https and runs on
port 9090.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,0 +1,10 @@
+FROM golang:latest
+LABEL maintainer="Mike Lloyd <mike@reboot3times.org>"
+
+RUN go get -v -d github.com/allegro/bigcache && \
+    go build -v -o bigcache github.com/allegro/bigcache/server && \
+    mv bigcache /usr/bin/bigcache
+
+EXPOSE 9090
+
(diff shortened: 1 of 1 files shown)


mxplusb:
Do not merge this yet, it's blocked on the SSL support.

server/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:latest
```
janisz: This is a build image, not a production one. It should be mentioned in a comment. Deployment image could be based on scratch image assuming bigcache will be built statically.

mxplusb: This is what the [Docker Hub](https://hub.docker.com/_/golang/) website says about that image:
```
This is the defacto image. If you are unsure about what your needs are, you probably want to use this one. It is designed to be used both as a throw away container (mount your source code and start the container to start your app), as well as the base to build other images off of.
```
I'm not sure I understand your feedback about the image base. If your concern is about the bigcache versioning within that image, we can tie it to a specific version. If you are referencing building an entirely custom container, that might be something to purse later, but I'm not looking to do that now since there are other things I'd rather focus on.

mxplusb:
Not sure when I'll get to this, closing for now.

PR #77, opened 2018-03-28
SSL Support

mxplusb:
This adds SSL support to the web server, and forces a user to opt out of it if they don't pass certificates.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -21,13 +21,17 @@ const (
 	statsPath = apiBasePath + "stats"
 
 	// server version.
-	version = "1.0.0"
+	version = "1.1.0"
 )
 
 var (
-	port    int
-	logfile string
-	ver     bool
+	port        int
+	logfile     string
+	ver         bool
+	noSSL       bool
+	defaultPort int = 443
+	serverCert  string
+	serverKey   string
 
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -41,9 +45,12 @@ func init() {
 	flag.DurationVar(&config.LifeWindow, "lifetime", 100000*100000*60, "Lifetime of each cache object.")
 	flag.IntVar(&config.HardMaxCacheSize, "max", 8192, "Maximum amount of data in the cache in MB.")
 	flag.IntVar(&config.MaxEntrySize, "maxShardEntrySize", 500, "The maximum size of each object stored in a shard. Used only in initial memory allocation.")
-	flag.IntVar(&port, "port", 9090, "The port to listen on.")
+	flag.IntVar(&port, "port", 443, "The port to listen on.")
```
janisz: Can we use `defaultPort` here? I'm not a big fan of changing the default behaviour. We need to add readme entry about this change and short instruction how to run cache.

mxplusb: Sure.

mxplusb:
Closing since I have no idea when I'll get to this.

PR #80, opened 2018-04-05
Add remove reason signal to OnRemove callback

jshufro:
I'm still pretty new to golang, so this PR probably has a C-like odor to it (especially with regard to the bitmasks). The general purpose is to indicate in the OnRemove callback _why_ a key is being removed. There are 3 reasons, represented as bitmasks by leftshifting iota in a const expression. I'm not sure if there's a more go-like approach, but I'd like to see support for this feature so I thought I'd open a PR and get a discussion going.

As for the API- I've made a small breaking change. Users of the library will have to add a parameter to their OnRemove callbacks, which can be ignored. If it's preferable to add new functions instead of changing the signature if existing ones, I'd be happy to adapt the code.


bigcache_test.go
```diff
@@ -203,6 +204,39 @@ func TestOnRemoveCallback(t *testing.T) {
...
+	onRemove := func(key string, entry []byte, reason RemoveReason) {
+		onRemoveInvoked = true
+	}
+	cache, _ := newBigCache(Config{
+		Shards:             1,
+		LifeWindow:         time.Second,
+		MaxEntriesInWindow: 1,
+		MaxEntrySize:       256,
+		OnRemove:           onRemove,
+		OnRemoveFilter:     Deleted | NoSpace,
```
mxplusb: Should this be `||` once you fix the bit shifting?

bigcache_test.go
```diff
@@ -203,6 +204,39 @@ func TestOnRemoveCallback(t *testing.T) {
 	assert.True(t, onRemoveInvoked)
 }
 
+func TestOnRemoveFilter(t *testing.T) {
```
mxplusb: Need to add another test that would cover `Expired` as well. Or implement `Expired` in this test. :)

mxplusb:
Also, Travis is failing due to formatting, so make sure to run `go fmt` on files you change so it doesn't error out on linting. 😄

bigcache.go
```diff
@@ -22,6 +22,18 @@ type BigCache struct {
 	maxShardSize uint32
 }
 
+// RemoveReason is a value used to signal to the user why a particular key was removed in the OnRemove callback.
+type RemoveReason int32
+const (
+	// Expired means the key is past its LifeWindow.
+	Expired    = 1 << iota
```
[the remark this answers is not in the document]
jshufro: from my understanding, what your suggesting will assign the sequence of values as 0, 1, 2

However, for bitmasks, the correct sequence is 1,2,4,8,16 etc
since the binary representation of 4 is, for example, 0100, you can combine these masks with the bitwise | to store multiple bools on the same integer. Thus, Expired | Deleted actually equals 0101, which is what lets me store just one integer instead of several bools

mxplusb:
I really like this change, I would posit there is a path forward via versioning:

Current: `OnRemove` + `OnRemoveExt`
Current+1: `OnRemove` with deprecation notice + `OnRemoveExt`
Current+2: `OnRemoveExt`

I do agree we shouldn't break backwards compatibility but this is a nice change.

jshufro:
I've made the recommended changes with regard to the API breakage. It should be easy to deprecate OnRemove when the time comes.

The only thing worth noting is that OnRemoveExt will be ignored as long as OnRemove is provided, as users shouldn't use both. I've updated the comments/docs to reflect this

mxplusb:
@jshufro can you please fix the bit shifting iotas? The rest looks good.  :)

jshufro:
@mxplusb I replied inline earlier - https://github.com/allegro/bigcache/pull/80#discussion_r179586768

README.md
```diff
@@ -46,10 +46,20 @@ config := bigcache.Config {
...
-		// means no callback and it prevents from unwrapping the oldest entry.
+		// callback fired when the oldest entry is removed because of its expiration time or no space left
+		// for the new entry, or because delete was called. A bitmask representing the reason will be returned.
+		// Default value is nil which means no callback and it prevents from unwrapping the oldest entry.
 		OnRemove: nil,
+		// callback fired when the oldest entry is removed because of its expiration time or no space left
+		// for the new entry, or because delete was called. A bitmask representing the reason will be passed through.
+		// Default value is nil which means no callback and it prevents from unwrapping the oldest entry.
+		// Ignored if OnRemove is specified.
+		OnRemoveExt func(key string, entry []byte, reason RemoveReason)
```
janisz: I'd prefer different name `OnRemoveWithReason`

bigcache.go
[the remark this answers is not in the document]
mxplusb: Mmmm while I do appreciate the optimisation, I am going to disagree and say while it is an integer optimization, the benefit of the optimisation underweighs against the idiotmatic use of `iota`. When you filed the PR, you made this statement:

> I'm still pretty new to golang, so this PR probably has a C-like odor to it

Bit shifting an `iota` is something you shouldn't really be doing in idiomatic Go as a lot of the reason for `iota` was to provide an enumerator that you don't have to manage or think about - the runtime manages it.

I'm going to defer to @janisz, but my stance is that bit shifting an `iota` is over-engineering and not idiomatic.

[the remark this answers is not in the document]
mxplusb: > In C we always avoid doing long, sequential boolean checks using bitmaps.

I agree, but this is Go, [you can always do](https://play.golang.org/p/paDVmrAH9rC):

```go
switch reason {
  case Expired: //
  case NoSpace: // 
  case Deleted: //
 }
```

Which is both cleaner logic as well as being idiomatic. More lines of code for idiomacy is better.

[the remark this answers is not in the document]
jshufro: The switch is definitely cleaner- but it still requires 3x varaibles on the config instead of one, for the filter

jshufro:
sorry for the radio silence- i'll get back to this shortly.

I actually have a decent idea of how to keep the single bitmask for the filter without shifting the constants

mxplusb:
@jshufro excited to see what you come up with!

jshufro:
I pushed up a change that does the following:
1) no longer bitshift by iota, but instead bitshift when reading/writing the filter mask
2) add a variadric function to set the filter mask
3) rename OnRemoveExt to OnRemoveWithReason
and obviously update docs/comments

bigcache.go
```diff
@@ -22,6 +22,19 @@ type BigCache struct {
 	maxShardSize uint32
 }
 
+// RemoveReason is a value used to signal to the user why a particular key was removed in the OnRemove callback.
+type RemoveReason int32
+
+const (
+	// Expired means the key is past its LifeWindow.
+	Expired = iota
```
cristaloleg: `Expired RemoveReason = iota`, without an explicit type it will be `int`.
See example https://play.golang.org/p/mZZdMaI92cI

bigcache_test.go
```diff
@@ -203,6 +238,40 @@ func TestOnRemoveCallback(t *testing.T) {
...
+	// when
+	cache.Set("key", []byte("value"))
+	clock.set(5)
+	cache.Set("key2", []byte("value2"))
+
+	// then
+	assert.False(t, onRemoveInvoked)
+
+	// and when
+	cache.Delete("key2")
```
cristaloleg: `onRemoveInvoked` should be set to `false` before this line, isn't it?

bigcache.go
jshufro: Neat. I'll update.

bigcache_test.go
jshufro: It's initialized to false on L246 and asserted to still be false on L266

I can explicitly set it to false here if it makes things clearer?

cristaloleg: Ah, indeed. So just ignore my comment, sorry :)

mxplusb:
@jshufro lgtm, thanks for the change!

jshufro:
Had to amend the commit to add a gpg signature

jshufro:
I signed this with a key for the wrong email. I'll resign it.

jshufro:
Fixed the gpg signature. I promise this is the last update

Issue #83, opened 2018-05-18
Feedback from Community Talk


mxplusb:
I totally forgot to capture this feedback live, and that's something I regret doing, so I will try to capture this well. I'm tagging the people I can remember being there, so if you see someone from the meetup is missing and you want to make sure their feedback is captured, please add them!

[Here](https://pivotal.zoom.us/recording/play/v7ijVuoFILrBPsSSqPH9ZM000f8k-3QpAZ0gRgn-LGH4_l3nOGz8Uq3Lqiwki3Q2) is the talk. The audio is sketchy for some reason, but next time I give this talk, I can make sure it's better.

* Everyone loved the performance focus!
* What are the use cases?
* Does the cache persist past program execution? i.e., can you restart the cache and have it rebuild itself
* Is there an option to distribute the cache? i.e., cache clustering
* How do we handle eviction?
* Is the cache thread safe?
* What is the performance like with billions of objects/hundreds of GBs?

Would love to really involve the community on this one!

//cc @enocom @jasonkeene @kcboyle

mfrw:
Would it be possible to make the talk accessible to all?

mxplusb:
I didn't realise it was auth locked, here is a Youtube version: https://youtu.be/K15B1FYkLDg. The audio is still sketchy because of the recording equipment.

mxplusb:
Closing since any feedback has been long captured.

PR #85, opened 2018-05-19
WIP: Persistence

mxplusb:
WIP: Added Export() to export a cache.

This is a work-in-progress for exporting the entire cache so it can be stored on disk. This layout/format isn't set in stone, nor is it terribly performant currently. Would love some feedback on how to improve the performance. I think it would be good to pipeline the data export a bit, so instead of trying to serialize the entire cache, you can pipeline it with a io.Writer and channels.


mxplusb:
There is still _a lot_ of work left to be done on this PR.

* I have to make sure that it can be done concurrently *and* asynchronously.
* I still have to implement `Import()`.
* We have to agree on the interface.
* I have to go through and figure out how to make the export performant.
* Data integrity is very important, so we definitely need a way to export it, but it should be _fast enough_, which is a problem when you go to checksum it.
* `Import()` should be _fast enough_.
* We have to agree on the file format.
* I need to figure out how to do this without copying the entire cache (thus doubling a memory requirement).

I wanted to push an early commit so I can get some early feedback. This directly relates to feedback in #83.

README.md
```diff
@@ -145,6 +145,17 @@ This package also includes an easily deployable HTTP implementation of BigCache,
 
 Bigcache genesis is described in allegro.tech blog post: [writing a very fast cache service in Go](http://allegro.tech/2016/03/writing-fast-cache-service-in-go.html)
 
+## Working with Protobufs
```
janisz: I'd prefer to not use protobuf. Let's stick with stdlib and/or add user wasy to use custom serialization

.travis.yml
```diff
@@ -1,31 +1,31 @@
...
-  - diff <(echo -n) <(gofmt -s -d .)
-  - golint ./... # This won't break the build, just show warnings
-  - ineffassign .
-  - go vet ./...
-  - go test -race -count=1 -coverprofile=queue.coverprofile ./queue
-  - go test -race -count=1 -coverprofile=server.coverprofile ./server
-  - go test -race -count=1 -coverprofile=main.coverprofile
-  - $HOME/gopath/bin/gover
-  - $HOME/gopath/bin/goveralls -coverprofile=gover.coverprofile -service travis-ci
+language: go
```
janisz: Revert changes

README.md
mxplusb: I agree that protobufs aren't ideal. I'm open to alternatives that you suggest. I tried playing around with `encoding/gob`, but it was really slow and I couldn't checksum shard instances. `encoding/json` isn't the right path since it's slower than `gob` and would have a complicated file structure, and you also can't checksum a shard instance.

We will also need an across-the-wire format, as well, if we want to have a discussion about a distributed cache as part of the feedback about #83. 

Like I mentioned, I'm open to alternatives that you want to suggest that are still performant and let us checksum shard instances.

cristaloleg: We can always start with a `encoding/gob` to have a proof of concept. /shrug

mxplusb:
Closing until I have a better solution in mind.

PR #86, opened 2018-05-19
Fix set benchmark to be more fair to maps

jasonkeene:
Big/free cache get hints as to how much data they will store. To be fair map should get the same hint so it doesn't have to pay the cost of growing over time.

This changed `BenchmarkMapSet` from:

```
BenchmarkMapSet-8      2000000      612 ns/op     336 B/op     3 allocs/op
```

to

```
BenchmarkMapSet-8      5000000      429 ns/op     210 B/op     3 allocs/op
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_install:
   - go get github.com/mattn/goveralls
   - go get golang.org/x/tools/cmd/cover
   - go get golang.org/x/tools/cmd/goimports
-  - go get github.com/golang/lint/golint
+  - go get golang.org/x/lint/golint
   - go get github.com/stretchr/testify/assert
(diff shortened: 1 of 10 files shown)


jasonkeene:
I signed this and rebased it onto https://github.com/allegro/bigcache/pull/84 to avoid a merge commit or a rebase that loses the signature.

mxplusb:
I'll merge this onto #84 once it's ready.

andersfylling:
Is this going to be updated? 

I really like bigcache, so I'd also hope to see correct perf numbers. Right now, it seems the README file is a little missleading.

Nice work btw ^^,

andersfylling:
Remember to update the README file ^^,

cristaloleg:
In progress 😉

PR #89, opened 2018-08-27
capacity field

### LINKED ISSUE #88: Expose ByteQueue capacity from Cache methods
The cache already exposes the number of entries in the cache through `Len()` and some additional useful statistics through `Stats()`. I would like to be able to get the underlying size in bytes of the cache, either from a method or as a new field of the `Stats` type.

The `ByteQueue` package already have a `Capacity()` methods that returns exactly the information I need. Could we somehow, using one of the two techniques before, expose the size of the cache ?

Thanks

cristaloleg:
Fixes #88

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -124,6 +124,15 @@ func (c *BigCache) Len() int {
 	return len
 }
 
+// Capacity returns amount of bytes store in the cache.
+func (c *BigCache) Capacity() int {
+	var len int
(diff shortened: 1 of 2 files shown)


wI2L:
@cristaloleg LGTM

cristaloleg:
@janisz it's a simple field exposing, 0 new logic.
Can we left tests for this change as a separate issue atm? :)

PR #106, opened 2018-11-13
Avoid data copying in Set()

MoZhonghua:
In current implemention, Set() first encode entry to local buffer, and 
then push data into queue. In this PR, Set() encodes entry directly into 
queue.

| name | old time/op | new time/op | delta |
| --- | --- | --- | --- |
| BigCacheSet-4 | 983ns ± 8% | 925ns ± 5% |-5.92% (p=0.040 n=5+5) |
| BigCacheSetParallel-4 | 521ns ± 4% | 496ns ± 4% |-4.80% (p=0.032 n=5+5) |

### DIFF
--- a/queue/bytes_queue_test.go
+++ b/queue/bytes_queue_test.go
@@ -26,6 +26,28 @@ func TestPushAndPop(t *testing.T) {
 	assert.Equal(t, entry, pop(queue))
 }
 
+func TestPushAndPopDirty(t *testing.T) {
+	t.Parallel()
+
+	// given
+	queue := NewBytesQueue(10, 0, true)
+	entry := []byte("hello")
+
+	// when
+	_, err := queue.Pop()
+
+	// then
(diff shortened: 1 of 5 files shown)


cristaloleg:
Thank you for contributing. I'll review it in next few days.

Can you fix a CI ? There is a problem with gofmt 😉

queue/bytes_queue_test.go
```diff
@@ -26,6 +26,28 @@ func TestPushAndPop(t *testing.T) {
 	assert.Equal(t, entry, pop(queue))
 }
 
+func TestPushAndPopDirty(t *testing.T) {
```
mxplusb: Can you please add a test which covers the queue max capacity? Thanks. :)

mxplusb:
I like the direction of this PR. I think `PushDirty` could use a bit more test coverage, but otherwise I think it's a good idea. 👍

mxplusb:
Closing due to age.

PR #115, opened 2019-01-16
Add invalid put test that make 500 error in server

u5surf:
#112 
I also added some tests for which increase test coverage, made it indicated to be internal server error when PUT request in server.go

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"io/ioutil"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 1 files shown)


server/server_test.go
```diff
@@ -247,3 +248,43 @@ func TestCacheIndexHandler(t *testing.T) {
...
+
+	putCacheHandler(rr, req)
+	resp := rr.Result()
+
+	if resp.StatusCode != 500 {
+		t.Errorf("want: 500; got: %d", resp.StatusCode)
+	}
+}
+
+func blob(char byte, len int) []byte {
```
cristaloleg: https://godoc.org/bytes#Repeat

server/server_test.go
```diff
@@ -247,3 +248,43 @@ func TestCacheIndexHandler(t *testing.T) {
...
+	b := make([]byte, len)
+	for index := range b {
+		b[index] = char
+	}
+	return b
+}
+
+type errReader int
+
+func (errReader) Read(p []byte) (n int, err error) {
```
cristaloleg: Due to dropped variable for `errReader` everything else can be also dropped 😉 
```suggestion
func (errReader) Read([]byte) (int, error) {
```

server/server_test.go
u5surf: awesome , I think all blob function should be replaced to bytes.Repeat.

mxplusb:
@u5surf can you please sign d42f76f or squash all of the commits so it can be merged? It looks good, just waiting on that.

Issue #117, opened 2019-02-11
Possible datarace


holiman:
I was looking into the `shard.go`, and saw the pattern below which looks racey to me. I haven't gone through all the steps to fuzz or repro it, but thought I'd post it here to get your thoughts about it. 
Possibly problematic code :

```golang
func (s *cacheShard) del(key string, hashedKey uint64) error {
	s.lock.RLock()
	itemIndex := s.hashmap[hashedKey]

	if itemIndex == 0 {
		s.lock.RUnlock()
		s.delmiss()
		return ErrEntryNotFound
	}

	wrappedEntry, err := s.entries.Get(int(itemIndex))
	if err != nil {
		s.lock.RUnlock()
		s.delmiss()
		return err
	}
	s.lock.RUnlock()

	s.lock.Lock()
	{
		delete(s.hashmap, hashedKey)
		s.onRemove(wrappedEntry, Deleted)
		resetKeyFromEntry(wrappedEntry)
	}
	s.lock.Unlock()

	s.delhit()
	return nil
}
```
Multiple readers may enter the first section in paralell, each getting an identical `wrappedEntry` for an entry to delete. They will then enter the writelocked section sequentially, and `resetKeyFromEntry` will write destructively on the data. It's possible that the sequential deletion of the same element is fine, but even so it might be that another write reuses the data, and the next of these sequential reads will overwrite the data. 

The proper pattern should be to either wrap everything within write-lock, or redo the first check once the writelock is obtained. 
```golang
func (s *cacheShard) del(key string, hashedKey uint64) error {
	s.lock.RLock()
	itemIndex := s.hashmap[hashedKey]

	if itemIndex == 0 {
		s.lock.RUnlock()
		s.delmiss()
		return ErrEntryNotFound
	}

	wrappedEntry, err := s.entries.Get(int(itemIndex))
	if err != nil {
		s.lock.RUnlock()
		s.delmiss()
		return err
	}
	s.lock.RUnlock()

	s.lock.Lock()
	{

		itemIndex = s.hashmap[hashedKey]

		if itemIndex == 0 {
			s.lock.Unlock()
			s.delmiss()
			return ErrEntryNotFound
		}

		wrappedEntry, err = s.entries.Get(int(itemIndex))
		if err != nil {
			s.lock.Unlock()
			s.delmiss()
			return err
		}

		delete(s.hashmap, hashedKey)
		s.onRemove(wrappedEntry, Deleted)
[...]

PR #119, opened 2019-02-15
Fix data corruption (shard.go)

holiman:
This PR contains a testcase which demonstrates corruption (intermittently). Sometimes leading to `panic`, and sometimes leading to data corruption of values. 

I thought that fixing the datarace suggested in #117 would solve it, but it seems I was wrong about that, there's some other underlying bug. I'll push a commit on top if I find it.

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -377,7 +377,7 @@ func TestCacheDelRandomly(t *testing.T) {
 	//c.Hasher = hashStub(5)
 	cache, _ := NewBigCache(c)
 	var wg sync.WaitGroup
-	var ntest = 200000
+	var ntest = 800000
 	wg.Add(1)
 	go func() {
 		for i := 0; i < ntest; i++ {
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -362,7 +362,7 @@ func TestCacheDel(t *testing.T) {
 
 // TestCacheDelRandomly does simultaneous deletes, puts and gets, to check for corruption errors.
 func TestCacheDelRandomly(t *testing.T) {
-
+	t.Parallel()
 	c := Config{
 		Shards:             1,
 		LifeWindow:         time.Second,
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -1,7 +1,9 @@
 package bigcache
 
 import (
+	"bytes"
 	"fmt"
+	"math/rand"
 	"runtime"
 	"sync"
 	"testing"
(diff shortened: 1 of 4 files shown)


holiman:
I pushed a commit on top which fixes a datarace in `shard.go`, where a slice was not copied into a new buffer before the lock was released. That fixes _most_ of the corruptions, but not all failures.

holiman:
Before the fix in `del` (7a2f9a270771b1b06b00b59f5844bd6705abc516), running with `ntests` `800000`, I got test failures `6` times out of `10`.

With the last fix, it passes `10` out of `10` runs.

bigcache_test.go
```diff
@@ -358,6 +360,67 @@ func TestCacheDel(t *testing.T) {
 	assert.Len(t, cachedValue, 0)
 }
 
+// TestCacheDelRandomly does simultaneous deletes, puts and gets, to check for corruption errors.
+func TestCacheDelRandomly(t *testing.T) {
+
```
mxplusb: Since this is standing up a separate cache, please call `t.Parallel()`.

queue/bytes_queue.go
```diff
@@ -162,6 +162,11 @@ func (q *BytesQueue) Get(index int) ([]byte, error) {
 	return data, err
 }
 
+// CheckGet checks if an entry can be read from index
+func (q *BytesQueue) CheckGet(index int) error {
```
mxplusb: Please add a test for this.

queue/bytes_queue.go
```diff
@@ -177,6 +182,23 @@ func (e *queueError) Error() string {
 	return e.message
 }
 
+// peekCheckErr is identical to peek, but does not actually return any data
+func (q *BytesQueue) peekCheckErr(index int) error {
```
mxplusb: Also needs a test. :)

holiman:
I addressed the concerns.

mxplusb:
So that's weird. I merged it in the UI, it didn't merge the branch, and then it just kicked off another travis build, but then travis errored out:

```bash
$HOME/gopath/bin/goveralls -coverprofile=gover.coverprofile -service travis-ci
Bad response status from coveralls: 422
{"message":"Couldn't find a repository matching this job.","error":true}
The command "$HOME/gopath/bin/goveralls -coverprofile=gover.coverprofile -service travis-ci" exited with 1.
```

@cristaloleg can you take a look when you get a chance?

cristaloleg:
Thank you @holiman

PR #120, opened 2019-02-16
add a fuzzer

holiman:
This PR adds a fuzzer. In case you don't want too long-running tests on CI, this can be run independently (say, a couple of hours) from time to time to get extensive coverage. 

On current master, it crashes pretty quickly, but have not yet crashed on https://github.com/allegro/bigcache/pull/119


holiman:
Hm, that looks odd. Running go format locally doesn't change anything. Not sure if it's a version compatibility between what I run and what you're using on travis, or if the travis script has some error in it. Feel free to push on top of my branch

PR #136, opened 2019-06-03
Hiqinternational patch 1

hiqinternational:
OK, luckily I have this "older" version. Is this ok? I can improve on this "branch"

### DIFF
--- a/fnv.go
+++ b/fnv.go
@@ -1,7 +1,7 @@
 package bigcache
 
 import(
-	"github.com/cespare/xxhash"
+	xxhash "github.com/cespare/xxhash"
 )
 // newDefaultHasher returns a new 64-bit FNV-1a Hasher which makes no memory allocations.
 // Its Sum64 method will lay the value out in big-endian byte order.
--- a/fnv.go
+++ b/fnv.go
@@ -21,6 +21,6 @@ const (
 */
 // Sum64 gets the string and returns its uint64 hash value.
 func (f fnv64a) Sum64(key string) uint64 {
-	hash := xxhash.Sum64(key)
+	hash := xxhash.Sum64([]byte(key))
 	return hash
 }
--- a/fnv.go
+++ b/fnv.go
@@ -1,32 +1,26 @@
 package bigcache
 
 import(
-	xxhash "github.com/cespare/xxhash"
+	"github.com/cespare/xxhash"
 )
 // newDefaultHasher returns a new 64-bit FNV-1a Hasher which makes no memory allocations.
 // Its Sum64 method will lay the value out in big-endian byte order.
 // See https://en.wikipedia.org/wiki/Fowler–Noll–Vo_hash_function
-
 func newDefaultHasher() Hasher {
 	return fnv64a{}
 }
 
 type fnv64a struct{}
-
+/*
 const (
 	// offset64 FNVa offset basis. See https://en.wikipedia.org/wiki/Fowler–Noll–Vo_hash_function#FNV-1a_hash
 	offset64 = 14695981039346656037
 	// prime64 FNVa prime value. See https://en.wikipedia.org/wiki/Fowler–Noll–Vo_hash_function#FNV-1a_hash
 	prime64 = 1099511628211
 )
+*/
 // Sum64 gets the string and returns its uint64 hash value.
-func (f fnv64a) Sum64(key string) uint64 {
-//	var hash uint64 = offset64
-	hash := xxhash.Sum64([]byte(key))
-//	for i := 0; i < len(key); i++ {
-//		hash ^= uint64(key[i])
-//		hash *= prime64
-//	}
-
+func (f fnv64a) Sum64(key []byte) uint64 {
+	hash := xxhash.Sum64(key)
 	return hash
 }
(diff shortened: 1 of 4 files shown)


hiqinternational:
First time github user, sorry.

fnv.go
```diff
@@ -8,21 +11,16 @@ func newDefaultHasher() Hasher {
...
+*/
 // Sum64 gets the string and returns its uint64 hash value.
 func (f fnv64a) Sum64(key string) uint64 {
-	var hash uint64 = offset64
-	for i := 0; i < len(key); i++ {
-		hash ^= uint64(key[i])
-		hash *= prime64
-	}
-
+	hash := xxhash.Sum64([]byte(key))
```
mxplusb: Please provide a detailed explanation as to why this is more appropriate than FNV hashing, and the problems it solves which the FNV hashing cannot. Looking at the library, it contains this statement: _This is a high-quality hashing algorithm that is much faster than anything in the Go standard library._ We aren't using a golang standard library hashing function, we have implemented our own.

fnv.go
```diff
@@ -8,21 +11,16 @@ func newDefaultHasher() Hasher {
 }
 
 type fnv64a struct{}
-
+/*
```
mxplusb: Please remove the comments.

fnv.go
```diff
@@ -1,5 +1,8 @@
 package bigcache
 
+import(
+	"github.com/cespare/xxhash"
+)
 // newDefaultHasher returns a new 64-bit FNV-1a Hasher which makes no memory allocations.
```
mxplusb: Please remove the comments.

mxplusb:
It looks like this patch has [linting failures](https://travis-ci.org/allegro/bigcache/jobs/540899811#L524) and [test failures](https://travis-ci.org/allegro/bigcache/jobs/540899811#L571). Please make sure to address those.

fnv.go
hiqinternational: 1. quality of xxhash is much higher than the rest and faster:
http://cyan4973.github.io/xxHash/
2. https://github.com/cespare/xxhash
i've used it and a few other golang repo also used it and it's actually very good and stable and reliable. the only thing i cant understand is "we only go with standard library... golang" that doesnt really mean anything because if this kind of rationale is needed means: we should all stick to map[][] and not have bigcache... since bigcache is not standard caching of map items etc.
3. xxhash is used in two other very good repo too:
https://github.com/VictoriaMetrics/fastcache
https://github.com/coocood/freecache
4. golang (language version) is going to upgrade anyway, so let's used the best (hash) for now.
means none of us really know whether next gc written's map[][] will improve better than what bigcache is offered now etc etc.

hiqinternational:
Is this ok?

mxplusb:
@cristaloleg & @janisz how do you feel about this patch? I think the idea of the patch is great, performance improvements are awesome, I'm just concerned about the 3rd party support. Any thoughts or commentary would be good.

fnv.go
janisz: Can you run our benchmarks?

fnv.go
```diff
@@ -1,28 +1,20 @@
 package bigcache
 
-// newDefaultHasher returns a new 64-bit FNV-1a Hasher which makes no memory allocations.
-// Its Sum64 method will lay the value out in big-endian byte order.
-// See https://en.wikipedia.org/wiki/Fowler–Noll–Vo_hash_function
+import (
+	"github.com/cespare/xxhash"
```
janisz: I like the idea of not having any external dependencies.

fnv_test.go
```diff
@@ -1,8 +1,9 @@
 package bigcache
 
 import (
-	"hash/fnv"
 	"testing"
+
+	"github.com/pierrec/xxHash/xxHash64"
```
cristaloleg: But in code there is another package 🤔 
https://github.com/allegro/bigcache/pull/136/files#diff-05f222304907b8cb5a43cc60c71cdf9dR4

janisz:
I think we should stick to our policy to not use external libs. We can add information about `xxhash` in read-me with performance comparison and instructions how to use it. `Hasher` is configurable and `xxhash` already implements it.

mxplusb:
Instead of saying no, let's see if we can generalise the hashing interface, then allow users to implement the interface, that way we can ship with a default implementation, and then users can implement their own.

I think that would solve @hiqinternational's desire to have something faster, and then our desire to stick to the stdlib. Thoughts?

janisz:
> Instead of saying no, let's see if we can generalise the hashing interface, then allow users to implement the interface, that way we can ship with a default implementation, and then users can implement their own.

It's already done. We have [`Hasher`](https://godoc.org/github.com/allegro/bigcache#Hasher) interface and custom hasher could be configured in [config](https://godoc.org/github.com/allegro/bigcache#Config)

mxplusb:
You're right, I missed it. @hiqinternational I think the best way forward with your desired goals are to implement a custom hasher using the [`Hasher`](https://godoc.org/github.com/allegro/bigcache#Hasher) interface, then you can use whichever library you want. I'm going to close this out in favour of that method. Please let us know if that doesn't fit your use case.

Thanks all!

PR #137, opened 2019-06-06
Go module support

maciej:
This commit adds Go modules[1] support.

I ran:
```sh
go mod init github.com/allegro/bigcache
go mod tidy
```
in the project directory

[1] https://github.com/golang/go/wiki/Modules

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,9 @@
+module github.com/allegro/bigcache
+
+go 1.12
+
+require (
+	github.com/cespare/xxhash v1.1.0 // indirect
+	github.com/coocood/freecache v1.1.0
+	github.com/stretchr/testify v1.3.0
+)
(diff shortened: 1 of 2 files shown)


cristaloleg:
Thank you for the contribution.
AFAIR we need to release Bigcache v2 with go.mod so no one one 1.x will get a broken build.
@mxplusb do you've any additional info?

maciej:
I believe go modules are backward compatible, so addition of `go.mod` should not break existing builds not based on modules.

cristaloleg:
There was a thread on chi router regarding go modules, and they complained
about it. Maybe I’m mixing 2 different stories into one.

mxplusb:
@maciej this is shipped in [2.0!](https://github.com/allegro/bigcache/releases/tag/v2.0.0)

maciej:
Great, thank you!

PR #154, opened 2019-08-23
Add memory usage to README

mingard:
This PR adds some basic information to explain the harmlessness of exponential memory usage stats that may otherwise be regarded as cause for concern.


cristaloleg:
@mingard thanks for the docs and debug!
@janisz @mxplusb please have a look 😉

Issue #163, opened 2019-09-21
Confusing "GC pause time" section


FrankReh:
The README.md, line 135, still says the GC pause for map is over 10 seconds but the table shows only 52 ms. This discrepancy probably came from #69. Perhaps this line in the README should say the pause used to be over 10 seconds?

The go GC certainly has gone through huge strides since go1.6 so GC pause times might not be a good measure of bigcache's worth any longer. The GC might still be spending precious cycles marking allocated heap memory, but that isn't evident in the pause time.

janisz:
I think we somehow missed this PR in master branch https://github.com/mxplusb/bigcache/commit/1dc8897498212e7bb27c747a04ef9c7dca5c4ab7

janisz:
https://github.com/allegro/bigcache/pull/165

PR #165, opened 2019-09-23
Update README

### LINKED ISSUE #59: GC Pause time for maps.

janisz:
Looking at #59, I wanted to do a quick once over of the comparisons in
1.10 so we can make sure our numbers are still accurate. When looking
through the GC code, I discovered we are using the wrong GC stat. Using
debug.GCStats.Pause[0] uses the most recent GC pause, which can be zero
(as I found out on Windows). We should be using total pause for all
collections, which is debug.GCStats.PauseTotal. That will show the total
length of time for all GC activities during our test, which more
accurately reflects the type of comparison we are trying to show.

Also, I updated the readme to reflect our GC comparison with go1.10.

fixes #59 #163 

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,8 +131,6 @@ GC pause for map:  52.9661ms
 
 Test shows how long are the GC pauses for caches filled with 20mln of entries.
 Bigcache and freecache have very similar GC pause time.
-It is clear that both reduce GC overhead in contrast to map
-which GC pause time took more than 10 seconds.
 
 ### Memory usage
 


janisz:
This commit was added to #69 after it was merged so it did not land on the master branch.

cristaloleg:
Thanks @janisz

mxplusb:
Ah that's weird. Well, glad it made it there, at least. It would be good to go through and do another run-through of the readme to make sure it's still current (including comparisons).

janisz:
I'll take care of re running integration tests

PR #177, opened 2019-10-05
Build on Github Actions

gnomeria:
In the `lint code` step, I made an adjustment to allow failed `goimports` command to not exit the step. I noticed that running it in travis also results in that behaviour even though it's set to fail. We could have in the future a makefile and a comment back on the commit to know what adjustments need to be done.


cristaloleg:
image
I don't understand why it's not showed here 🤔

Issue #175, opened 2019-10-04
Migrate from Travis to Github Actions


cristaloleg:
Migrations should be quite easy, it should be enough to reimplement same steps in travis.yml to Github Actions config.

PLEASE if you're going to start working on this issue, comment below, to discuss the solution, thanks in advance.

gnomeria:
Hello, I'm interested on this issue. Can I do it? I'll start by reimplement from the travis.yml

cristaloleg:
@gnomeria yep, good luck

gnomeria:
I was having an issue with uploading the cover results to the coverall using goverall. Do you think we could switch to Codedov @cristaloleg ?

cristaloleg:
@gnomeria yeah, coverall returns 5xx last weeks as I see (other projects too). I'll make an issue regarding migration (let's make it in separate PRs, please)

cristaloleg:
Fixed in #180

PR #180, opened 2019-10-05
Migrate from travis to GitHub actions

gnomeria:
This will delete and disable the travis config hence its build checks for #175


janisz:
Fixed by https://github.com/allegro/bigcache/pull/179

PR #168, opened 2019-09-23
New Get API which returns additional information about the entry

jgheewala:
enhancement: added new data struct Response, which will be returned when
calling GetWithInfo(). Caller can use this Response to check the status
of the Entry, whether its Expired or not. Default '0' means its still
in use.

minor lint changes

Test: added test case for the api and made sure all tests cases are
successful

PR #181, opened 2019-10-09
Go defaults to "0"" so in case we want to return

jgheewala:
EntryStatus back to the caller "Expired" cannot be differentiated.
Fixing this by default "_" to 0 and incremental RemoveReasons

comment was missing for GetWithInfo api so updated that

test: ran all unit test cases


mxplusb:
I quite like this PR and problem it solves, but I agree this is definitely a 3.0 decision since it could break a lot of downstream configurations when they upgrade. This should definitely go in the release notes.

jgheewala:
@mxplusb this PR can be open until its ready to go in. I kept it here so that I do not forget about this :)

mxplusb:
Great, thanks!

flisky:
So, the error `ErrEntryIsDead` has no more referenced and should be removed ? Will give users a compile error in case they don't notice this breaking change.

cristaloleg:
@flisky good point, we can remove it. But anyway, this PR is for a next major version, so it will be safe to make a breaking change.

mxplusb:
@jgheewala when you get a chance, can you rebase? It would be good to keep this updated to prep for the next major version.

flisky:
May I take over of it, @jgheewala & @mxplusb ?

Besides these changes, I also want to return the expired entry even when `resp.EntryStatus = Expired`( `entry` goes `[]byte` when introduced in #168 , and then goes to `nil` in #191 ). Do you like the idea?

cristaloleg:
@flisky let's create a separate issue for your idea and discuss it there. This will make discussion much more visible and structured, thanks :)

jgheewala:
@mxplusb was super busy will rebase my changes soon

jgheewala:
@mxplusb done with the conflicts.

mxplusb:
@jgheewala there's a new branch, called `versions/v3-prep` which has the underlying necessary module changes to support this PR. If you can rebase against that branch, and then edit the PR to point to the `versions/v3-prep` branch, I can get this merged and we can close it out.

jgheewala:
@mxplusb https://github.com/allegro/bigcache/pull/255

janisz (approved):

cristaloleg (approved):

PR #184, opened 2019-11-04
Add entry stats

### LINKED ISSUE #159: Entry stats
Being able to see the number of times a cached resource was requested is a useful metric, especially at the point of removal.

As a production example, `OnRemove` could be used to trigger cache pre-warming based on the frequency of cache hits on a resource within a given window.

I'm aware of the overhead of storing such data by default and suggest this only as an optional parameter.

Thoughts?

Nazdroth:
This pull request resolves #159.

- Adding a way to access entry-level stats, that can be enabled/disabled using the config attribute `StatsEnabled` which is set to false by default.
- Adding a way to get metadata related to a particular key using `KeyMetaData(key string)`
-  OnRemove now can give you the possibility to access the entry metadata using `OnRemoveWithMetaData`. 
- Writing unit tests for the added functions.


mxplusb:
@cristaloleg do you know what happened to the checks?

config.go
```diff
@@ -28,7 +30,12 @@ type Config struct {
 	// OnRemove is a callback fired when the oldest entry is removed because of its expiration time or no space left
 	// for the new entry, or because delete was called.
 	// Default value is nil which means no callback and it prevents from unwrapping the oldest entry.
+	// ignored if OnRemoveWithMetaData is specified.
 	OnRemove func(key string, entry []byte)
+	// OnRemoveWithMetaData is a callback fired when the oldest entry is removed because of its expiration time or no space left
+	// for the new entry, or because delete was called. A structure representing details about that specific entry.
+	// Default value is nil which means no callback and it prevents from unwrapping the oldest entry.
+	OnRemoveWithMetaData func(key string, entry []byte, keyMetaData metaData)
```
mxplusb: Can you change this to `OnRemoveWithMetadata`? _Metadata_ doesn't have a capital _d_.

config.go
```diff
@@ -51,6 +58,7 @@ func DefaultConfig(eviction time.Duration) Config {
 		CleanWindow:        1 * time.Second,
 		MaxEntriesInWindow: 1000 * 10 * 60,
 		MaxEntrySize:       500,
+		StatsEnabled:       false,
```
mxplusb: I like that it's disabled by default.

shard.go
```diff
@@ -10,19 +10,25 @@ import (
 
 type onRemoveCallback func(wrappedEntry []byte, reason RemoveReason)
 
+type metaData struct {
```
mxplusb: Can you change this to `metadata`? _Metadata_ doesn't have a capital _d_.

shard.go
```diff
@@ -227,6 +238,12 @@ func (s *cacheShard) removeOldestEntry(reason RemoveReason) error {
 		hash := readHashFromEntry(oldest)
 		delete(s.hashmap, hash)
 		s.onRemove(oldest, reason)
+		if s.statsEnabled {
```
mxplusb: Why is this duplicated?

shard.go
```diff
@@ -265,8 +282,26 @@ func (s *cacheShard) getStats() Stats {
 	return stats
 }
 
-func (s *cacheShard) hit() {
+func (s *cacheShard) getKeyMetaData(key uint64) metaData {
```
mxplusb: Can you change this to `getKeyMetadata`? _Metadata_ doesn't have a capital _d_.

shard.go
```diff
@@ -10,19 +10,25 @@ import (
...
-	clock      clock
-	lifeWindow uint64
+	isVerbose    bool
+	statsEnabled bool
+	logger       Logger
+	clock        clock
+	lifeWindow   uint64
 
-	stats Stats
+	hashmapStats map[uint64]*uint32
```
cristaloleg: `uint32` should be enough here

shard.go
```diff
@@ -10,19 +10,25 @@ import (
 
 type onRemoveCallback func(wrappedEntry []byte, reason RemoveReason)
 
+type metadata struct {
```
cristaloleg: Why unexported? Same question about single field.

shard.go
Nazdroth: I used *uint32 here because you can't do `atomic.AddUint32(&s.hashmapStats[key], 1)` with a map

shard.go
Nazdroth: I followed the same logic as `cacheShard` but I can change that if you want.

cristaloleg: That's a big difference. We're exposing `metadata` type to the user (from `BigCache` method, and exposing unexported type is a strange thing.

shard.go
cristaloleg: We don't need atomic at all, 'cause all the write happens under a mutex.

shard.go
Nazdroth: should be fixed now

cristaloleg: Please, expose `requestCount` too (should be `RequestCount`) 😉

Nazdroth: done :+1:

Issue #189, opened 2019-12-02
Remove dependecy to github.com/stretchr/testify


janisz:
We are using only a small subset of github.com/stretchr/testify we can simply remove this dependecy and use `assertEqual` function. This will make our project dependency less.

https://github.com/allegro/bigcache/blob/master/go.mod#L5

mxplusb:
Yeah, I can agree to this. Might as well use only the standard library whenever possible.

PR #190, opened 2019-12-02
Remove dependecy to testify

janisz:
Fixes: #189

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -3,20 +3,16 @@ package bigcache
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"math/rand"
 	"runtime"
 	"sync"
 	"testing"
 	"time"
-
-	"github.com/stretchr/testify/assert"
 )
-
-var sink []byte
 
 func TestParallel(t *testing.T) {
 	t.Parallel()
-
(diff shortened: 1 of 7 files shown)


bigcache_test.go
```diff
@@ -34,7 +30,7 @@ func TestParallel(t *testing.T) {
 	go func() {
 		defer wg.Done()
 		for i := 0; i < keys; i++ {
-			sink, _ = cache.Get(fmt.Sprintf("key%d", i))
+			_, _ = cache.Get(fmt.Sprintf("key%d", i))
```
janisz: @cristaloleg I removed `sink`. Why did we needed this?
BTW: In this test there are no assertion. Is this right?

cristaloleg: `sink` was made to ensure that compiler will not remove operation as no-op.

janisz: How about deleteing this test? I think it duplicates `TestCacheDelRandomly`

https://github.com/allegro/bigcache/blob/ad8f119063878734a3466f293966e9da172e0213/bigcache_test.go#L438-L498

janisz:
@cristaloleg Can you review?

PR #191, opened 2019-12-02
Fix issue with concurrent read/writes

janisz:
Optimistic locking is not working for `getWrappedEntry` because it does not return copy of data so another thread can alter the data causing panic e.g., `panic: runtime error: slice bounds out of range [:57329] with capacity 20390`. To prevent this we need to acquire lock before getting wrapped entry and release it after all operations are done.

This bug was introduced in https://github.com/allegro/bigcache/commit/0bb712de7d60f23379d438fca050e9cdeac96c91

See: 

* https://github.com/janisz/bigcache/commit/e73c96d8b9c30e2f5247ce96521edbd62b5d096b/checks
* https://storytime.dev/preview/5de55ab8ddbb4e10b45671df

### DIFF
--- a/shard.go
+++ b/shard.go
@@ -38,15 +38,15 @@ func (s *cacheShard) getWithInfo(key string, hashedKey uint64) (entry []byte, re
 	wrappedEntry, err := s.getWrappedEntry(hashedKey)
 	if err != nil {
 		s.lock.RUnlock()
-		return wrappedEntry, resp, err
+		return nil, resp, err
 	}
 	if entryKey := readKeyFromEntry(wrappedEntry); key != entryKey {
+		s.lock.RUnlock()
+		s.collision()
 		if s.isVerbose {
 			s.logger.Printf("Collision detected. Both %q and %q have the same hash %x", key, entryKey, hashedKey)
 		}
-		s.lock.RUnlock()
-		s.collision()
-		return entry, resp, ErrEntryNotFound
+		return nil, resp, ErrEntryNotFound
 	}
 
 	oldestTimeStamp := readTimestampFromEntry(wrappedEntry)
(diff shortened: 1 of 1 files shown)


janisz:
@cristaloleg @mxplusb @jgheewala Can you take a look?

shard.go
```diff
@@ -34,68 +34,65 @@ type cacheShard struct {
...
-			s.lock.RUnlock()
-			// @TODO: when Expired is non-default value return err as nil as the resp will have proper entry status
-			resp.EntryStatus = Expired
-			return entry, resp, ErrEntryIsDead
-		}
-		entry := readEntry(wrappedEntry)
+	oldestTimeStamp := readTimestampFromEntry(wrappedEntry)
+	if currentTime-oldestTimeStamp >= s.lifeWindow {
 		s.lock.RUnlock()
-		s.hit(hashedKey)
```
jgheewala: @janisz I think we should still count it as hit, because the entry is still present in the cache after lifewindow expiry. If we do not count hit that would mis-lead to user who is looking at the stats for entries present in cache but not deleted.

janisz: I did not change this part.This is line 61 of the new code

janisz:
@cristaloleg @mxplusb @jgheewala Can you re-review? I added small improvements to return `nil` on error and to release lock earlier so logging is after critical section.

shard.go
jgheewala: 👍🏽

janisz:
@cristaloleg I lost my power to merge :( Can you merge this for me?

mxplusb:
Sorry, holiday week in the US recently so I've been out of reach. Looking at it, it looks okay to me. Are there any parts of bigcache which are still optimistic?

janisz:
> Are there any parts of bigcache which are still optimistic?
I don't know about any other.

PR #196, opened 2020-01-04
Add testable examples

janisz:
Copy examples from [#usage](https://github.com/allegro/bigcache#usage) into testable example.

See: https://blog.golang.org/examples

### DIFF
--- a/examples_test.go
+++ b/examples_test.go
@@ -1,0 +1,78 @@
+package bigcache_test
+
+import (
+	"fmt"
+	"log"
+	"time"
+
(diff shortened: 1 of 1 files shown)


examples_test.go
```diff
@@ -0,0 +1,78 @@
...
+		log.Fatal(err)
+	}
+
+	entry, err := cache.Get("my-unique-key")
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(string(entry))
+	// Output: value
+}
```
cristaloleg:
```suggestion
}

```

PR #197, opened 2020-01-07
Replace duplicated code with function call

janisz:
`peek` performs the same assertions as `peekCheckErr`. There is no need to keep this checks in two places.


mxplusb:
love it.

PR #199, opened 2020-01-08
Add test for parallel reads with stats

janisz:
Regresion test for https://github.com/allegro/bigcache/pull/198


mxplusb:
Love it.

mxplusb:
@cristaloleg now that #198 is merged, should a new release be cut so upstream consumers can get the fix?

PR #200, opened 2020-01-09
Move getOldestEntry to test file

janisz:
Move `getOldestEntry` from production code to test where it's used.

### DIFF
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -95,9 +95,15 @@ func TestEntriesIteratorWithConcurrentUpdate(t *testing.T) {
 		t.Errorf("Iterator should contain at least single element")
 	}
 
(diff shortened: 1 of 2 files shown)


mxplusb:
> I'm not a big fan of unneede closure

@cristaloleg would you be able to clarify what you mean by that? Are you saying that the whole lifecycle of this PR was too fast? It was a simple change and all the tests passed, so I didn't see a reason to wait on it since I was already looking at it.

cristaloleg:
@mxplusb No, everything is ok, just I'm trying to reduce amount of closures inside hot path when it's possible. AFAIR they're not inlined at all, which is a bit uncool for us.

mxplusb:
Gotcha, that makes sense. It's part of the test so I'm less concerned in this specific base, but I do agree that closures should be used carefully.

PR #203, opened 2020-02-04
Implemented new Append() method

### LINKED ISSUE #158: Implement new append function to use bigcache as index for internal database
Hi there,

I want to use bigcache for an index for my own "memory database".

My records have various fields and are indexed by a uint64 primary key. I now want to add secondary indices for columns pointing to primary keys. The data structure would look like the following:

bigcache[<columnvalue>] = `<uint64>,<uint64>,<uint64>`...

So for example my record has a column "city" and I want to an index on this column:

bigcache_city_index[newyork] = `<pk1>,<pk2>,<pk3>`...

From my point of view this cannot be done today because I will have a race condition between get() and set() calls. 

Right now I am thinking about the following:

- Implement new "AppendUint64(key, id)"
- Look up shard by key
- Lock shard
- Get value
- Add new id (uint64) to value (binary.PutUvarint())
- Set value
- Unlock shard

What are your thoughts about this?

Thank you,
Stefan

snacker81:
Implemented new Append() method with proper locking. Without this method you would need to wrap a Get()/Set() part with your own locks hurting performance.

Fixes #158


shard.go
```diff
@@ -75,16 +87,42 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
...
+	var newEntry []byte
+	oldEntry, err := s.get(key, hashedKey, true)
+	if err == nil {
+		newEntry = oldEntry
+	} else if err != ErrEntryNotFound {
+		s.lock.Unlock()
+		return err
+	} else {
+		// Entry not found
+	}
```
cristaloleg: I suppose this looks a bit more easy to follow, what do you think?
```
if err != nil {
	if err != ErrEntryNotFound {
		s.lock.Unlock()
		return err
	}
} else {
	newEntry = oldEntry
}
```

shard.go
```diff
@@ -25,40 +25,52 @@ type cacheShard struct {
 	stats Stats
 }
 
-func (s *cacheShard) get(key string, hashedKey uint64) ([]byte, error) {
-	s.lock.RLock()
+func (s *cacheShard) get(key string, hashedKey uint64, skipLock bool) ([]byte, error) {
```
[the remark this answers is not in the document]
cristaloleg: Ah, and the same regarding `set` method 😉

shard.go
```diff
@@ -75,16 +87,42 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
...
+		// Entry not found
+	}
+
+	newEntry = append(newEntry, entry...)
+	err = s.set(key, hashedKey, newEntry, true)
+	s.lock.Unlock()
+	if err != nil {
+		return err
+	}
+	return nil
```
cristaloleg: Probably this can be flatten to `return err` but it's a minor one.

snacker81: changed it, you are the boss ;)

shard.go
snacker81: yes changed it

shard.go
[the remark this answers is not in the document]
snacker81: I am really missing parameter default values here (which would be false of cause to not change the API).

But I have no better idea. I could clone the method but this will result in a "lot" of duplicate code. I would basically implement the method again just leaving out the lock lines?

Since those calls are not public I think we can live with it. What do you think?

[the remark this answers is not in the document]
cristaloleg: There is no need for a default params. Both of the functions are from different worls - one is lock-free and another is not.

Combining them is a way to make an error quite easily.

That's why I suggest to split them into 2 different. Little code duplication isn't a problem (especially here).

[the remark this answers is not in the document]
snacker81: done, good? :)

cristaloleg:
@janisz @mxplusb ?

cristaloleg:
btw, looks like your branch is outdated and you've a conflict, can you fix it too? pretty please :)

snacker81:
Fixed it, kinda "redid" the set()/get() part because I didnot want to screw up (merge conflict). Please check again carefully, thanks!

bigcache.go
```diff
@@ -134,6 +134,13 @@ func (c *BigCache) Set(key string, entry []byte) error {
 	return shard.set(key, hashedKey, entry)
 }
 
+// Append appends entry under existing key
```
janisz: Can you elaborate this? What if key does not exist and how it differs from `Set`

bigcache_test.go
```diff
@@ -27,6 +27,23 @@ func TestWriteAndGetOnCache(t *testing.T) {
...
+func TestAppendAndGetOnCache(t *testing.T) {
+	t.Parallel()
+
+	// given
+	cache, _ := NewBigCache(DefaultConfig(5 * time.Second))
+	value := []byte("value")
+
+	// when
+	cache.Append("key", value)
+	cache.Append("key", value)
```
janisz: We need a test to append different value and to run multiple appends in parallel.

shard.go
```diff
@@ -84,6 +84,24 @@ func (s *cacheShard) get(key string, hashedKey uint64) ([]byte, error) {
...
+		if s.isVerbose {
+			s.logger.Printf("Collision detected. Both %q and %q have the same hash %x", key, entryKey, hashedKey)
+		}
+		return nil, ErrEntryNotFound
+	}
+	entry := readEntry(wrappedEntry)
+	s.hit(hashedKey)
+
+	return entry, nil
+}
```
janisz: We have similar code in `get` and `getWithInfo` can we unify it?

shard.go
```diff
@@ -131,6 +149,51 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
 	}
 }
 
+func (s *cacheShard) setWithoutLock(key string, hashedKey uint64, entry []byte) error {
```
janisz: Can we use this function in `set`? After release of 1.14 we can loose out rule of no defer

shard.go
```diff
@@ -131,6 +149,51 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
...
+		}
+	}
+}
+
+func (s *cacheShard) append(key string, hashedKey uint64, entry []byte) error {
+	s.lock.Lock()
+	var newEntry []byte
+	oldEntry, err := s.getWithoutLock(key, hashedKey)
+	if err != nil {
+		if err != ErrEntryNotFound {
```
janisz: Can you add a test for this condition?

bigcache.go
snacker81: pls check

bigcache_test.go
snacker81: improved the test, the test is already marked with t.Parallel() is that what you mean?

shard.go
snacker81: ErrEntryNotFound is checked in the unit test with the first call to Append(). Or do you mean the error condition other than ErrEntryNotFound? How would I do that? :)

shard.go
snacker81: @cristaloleg this is not what you wanted right?

bigcache_test.go
```diff
@@ -27,6 +27,55 @@ func TestWriteAndGetOnCache(t *testing.T) {
 	assertEqual(t, value, cachedValue)
 }
 
+func TestAppendAndGetOnCache(t *testing.T) {
+	t.Parallel()
+
+	// given
+	cache, _ := NewBigCache(DefaultConfig(5 * time.Second))
+	key := "key"
+	value1 :=  make([]byte, 50)
```
janisz:
```suggestion
	value1 :=  blob('a', 50)
```

shard.go
snacker81: Not part of this pull request? :)

bigcache_test.go
```diff
@@ -27,6 +27,55 @@ func TestWriteAndGetOnCache(t *testing.T) {
...
+	rand.Read(value1)
+	value2 :=  make([]byte, 50)
```
janisz:
```suggestion
	value2 :=  blob('b', 60)
```

shard.go
cristaloleg: I suppose @janisz has proposed to use setWithoutLock inside a set. I think we can skip it for a moment and figure out that everything works (in a next PR we can integrate one into another). So if @janisz is agree - we can treat this comment as a resolved one.

bigcache_test.go
janisz: I thought about doing sometehing like we did here https://github.com/allegro/bigcache/blob/master/bigcache_test.go#L405-L406 just to check if locking is done properly.

shard.go
janisz: OK, let's stick with `no defer`. We need to discuss it.

bigcache_test.go
snacker81: ok good call, I implemented something like you mentioned and found another bug :)

One problem remains: the race detector triggers but I have no idea why. Do you guys have experience with it? Is it "valid" all the time?

snacker81: Ok fixed it, it was because I was using the "err" variable from the outer scope! :)

user3415653:
+1, need this for my project

cristaloleg:
@user3415653 can you provide more details? Why do you need this feature? (it's ready to merge, but I'm curious about a bit wider context, probably we can improve bigcache in the future). Thanks in advance.

cristaloleg:
@janisz PTAL 😉

cristaloleg:
Thanks @snacker81

mxplusb:
@cristaloleg I think it's too late now, but this should've been a v2.2.0 bump instead of v2.1.7 as it adds a new API.

cristaloleg:
@mxplusb good point! will publish v2.2.0

snacker81:
> Thanks @snacker81

@cristaloleg No problem thanks for the solution-oriented process :)

PR #210, opened 2020-03-19
Update Go version for CI

janisz:
Go 1.14 is around for some time


cristaloleg:
We can easily drop 1.12.x

janisz:
@cristaloleg done, please reconfigure required checks to drop 1.12

PR #211, opened 2020-03-20
[fix] set full to false in Reset()

neal-zhu:
Sorry about forgetting to reset ``q.full``

### DIFF
--- a/queue/bytes_queue.go
+++ b/queue/bytes_queue.go
@@ -8,7 +8,9 @@ import (
 
 const (
 	// Number of bytes used to keep information about entry size
-	headerEntrySize = 4
+	//headerEntrySize = 4
+	// Number of bytes to encode 0 in varint format
+	minimumHeaderSize = 1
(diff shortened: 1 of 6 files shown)


queue/bytes_queue.go
```diff
@@ -78,6 +78,7 @@ func (q *BytesQueue) Reset() {
 	q.head = leftMarginIndex
 	q.rightMargin = leftMarginIndex
 	q.count = 0
+	q.full = false
```
janisz: Can you add a test for this?

neal-zhu: I'd like to, but what kind of test do you mean? 
I think it's kind of obvious that without resetting this flag, the queue will still be full after ``Reset()``, and the next ``Push()`` will cause a re-allocation on this empty queue

janisz: Exactly,

neal-zhu:
Reopen a new one

janisz:
https://github.com/allegro/bigcache/pull/213

PR #216, opened 2020-04-01
give a way to access the stale cache

flisky:
an update for #181 :)

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -32,9 +32,9 @@ type Response struct {
 type RemoveReason uint32
 
 const (
+	_ RemoveReason = iota
 	// Expired means the key is past its LifeWindow.
-	// @TODO: Go defaults to 0 so in case we want to return EntryStatus back to the caller Expired cannot be differentiated
-	Expired RemoveReason = iota
+	Expired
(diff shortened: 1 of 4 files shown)


bigcache.go
```diff
@@ -32,9 +32,9 @@ type Response struct {
 type RemoveReason uint32
 
 const (
+	_ RemoveReason = iota
 	// Expired means the key is past its LifeWindow.
-	// @TODO: Go defaults to 0 so in case we want to return EntryStatus back to the caller Expired cannot be differentiated
-	Expired RemoveReason = iota
+	Expired
```
cristaloleg, line 37: looks like a breaking change

flisky: yes, it's a breaking change. The issue is discussed in #181, and hopefully in v3.0

mxplusb: I think I would like to see this explicitly defined as a numerical instead of an iota. That forces the API to remain stable, in my opinion.

flisky: @mxplusb updated, please take a look.

mxplusb:
@cristaloleg this looks good to me. I think it would be good to version bump after merging this because it sets an explicit value, so it's technically a minor API change.

@flisky can you please rebase and then re-request a review?

PR #218, opened 2020-04-29
fix: can not resuce bytes queue

mangotree2:
fix this:
`
// given
	cache, _ := bigcache.NewBigCache(bigcache.Config{
		Shards:             1,
		LifeWindow:         5 * time.Second,
		MaxEntriesInWindow: 2,
		MaxEntrySize:       400,
		HardMaxCacheSize:   1,
		Verbose:true,
	})

	// when
	_ = cache.Set("key1", blob('a', 1024*400))
	_ = cache.Set("key2", blob('b', 1024*400))
	_ = cache.Set("key3", blob('c', 1024*400))
	key4Err := cache.Set("key4", blob('d', 1024*400))
	if key4Err != nil {
		panic(key4Err)
	}`

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -747,6 +747,46 @@ func TestOldestEntryDeletionWhenMaxCacheSizeIsReached(t *testing.T) {
 	assertEqual(t, blob('c', 1024*800), entry3)
 }
 
+func TestNewEntryWriteWhenMaxCacheSizeIsReached(t *testing.T) {
+	t.Parallel()
+
+	// given
+	cache, _ := NewBigCache(Config{
+		Shards:             1,
+		LifeWindow:         5 * time.Second,
+		MaxEntriesInWindow: 2,
+		MaxEntrySize:       400,
+		HardMaxCacheSize:   1,
+	})
+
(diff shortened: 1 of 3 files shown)


mxplusb:
@mangotree2 would you be able to provide more context on what you're trying to fix? I don't understand the context of the problem nor the context of the proposed fix.

mangotree2:
@mxplusb  @neal-zhu 
 When maxCapacity is set and the queue is full, head will be equal to tail, and full will be set to true. When Push, canInsertAfterTail and canInsertBeforeHead will judge that the queue is full, it will remove the entry, but will not reset the full field, resulting in After the entry is removed, there is no room. 
Can run my test case，but I can’t modify it correctly - -

mxplusb:
I'm going to convert this to a draft so you can continue working on it until the tests pass. :)

cristaloleg:
@mangotree2 kindly ping 😉

mangotree2:
> @mangotree2 kindly ping 😉
-----
555 T.T .

cristaloleg:
Sorry, what?

Issue #126, opened 2019-05-15
OnRemoveWithReason still receives deleted entry


santatic:
I deleted a key from cache, `OnRemoveWithReason` receives entry with reason `bigcache.Deleted`
But then, at expire time  `OnRemoveWithReason` still receives deleted entry with reason `bigcache.Expired`

My config:
```go
var cache, _ = bigcache.NewBigCache(bigcache.Config{
		Shards:             16,
		LifeWindow:         time.Minute,
		CleanWindow:        500 * time.Millisecond,
		MaxEntriesInWindow: 5,
		MaxEntrySize:       1 * 1024 * 1024,
		Verbose:            true,
		HardMaxCacheSize:   5,
		OnRemove:           nil,
		OnRemoveWithReason: myRemoveWithReasonFunc,
		Logger:             bigcache.DefaultLogger(),
	})
``` 

- My OS: Ubuntu 18.04
- My go version: go1.12.2 linux/amd64
- My bigcache version: v1.2.0

mxplusb:
> at expire time OnRemoveWithReason still receives deleted entry with reason bigcache.Expired

@santatic are you saying the entry isn't actually being deleted? Can you explain the problem/question a bit more?

santatic:
> @santatic are you saying the entry isn't actually being deleted?

Yes, that's my case.

My steps:
1. I add key `a` = `v`
2. I delete key `a`, `OnRemoveWithReason` response key `a` with reason `bigcache.Deleted`
3. I wait for expire time `LifeWindow`, `OnRemoveWithReason` response key `a` with reason `bigcache.Expired`

mxplusb:
Let me see if I can reproduce it with a test case and if I can, I'll get it fixed, if not I'll have to ask you for a reproduction so that we can ensure there are no regressions later.

ilyaglow:
Here's the test that catches this case:
```diff
diff --git a/bigcache_test.go b/bigcache_test.go
index 580777d..2444dd5 100644
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -275,6 +275,55 @@ func TestOnRemoveFilter(t *testing.T) {
        assert.True(t, onRemoveInvoked)
 }

+func TestOnRemoveFilterExpired(t *testing.T) {
+       t.Parallel()
+
+       // given
+       clock := mockedClock{value: 0}
+       onDeleteInvoked := false
+       onExpireInvoked := false
[...]

PR #219, opened 2020-05-04
Deleted entries should not be removed by expire eviction

bachhh:
Hi, this is a small fix for #126,

I think the problem is that `shard.delete()` marks entries in queue as deleted ( by zero-ing all the 64 header bits for hash ), but `removeOldestEntry` is not checking for this case.

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -353,6 +353,67 @@ func TestOnRemoveFilter(t *testing.T) {
 	assertEqual(t, true, onRemoveInvoked)
 }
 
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,3 +1,5 @@
-module github.com/allegro/bigcache/v2
+module bigcache
```
cristaloleg, line 1: you shouldn't change this file 😄

bachhh: oh sorry, I accidentally pushed local changes to my master,

mxplusb:
Fixes #126.

Issue #214, opened 2020-03-28
Set failure even when there is enough memory left and size of entry is acceptable


paritosh-desai:
**What is the issue you are having?**
Set failing with `entry is bigger than max shard size` even when there is enough memory left and size of entry is acceptable

**What is BigCache doing that it shouldn't?**
[Error in question](https://github.com/allegro/bigcache/blob/master/shard.go#L147)
Reason [removeOldestEntry](https://github.com/allegro/bigcache/blob/master/shard.go#L145) returned an error 
Which was in-turn returned by[Pop](https://github.com/allegro/bigcache/blob/master/shard.go#L292)
[Pop](https://github.com/allegro/bigcache/blob/master/queue/bytes_queue.go#L159) will only return an error if `peek` fails 
[Peek](https://github.com/allegro/bigcache/blob/master/queue/bytes_queue.go#L229) peek `can` return an error if the [queue is empty](https://github.com/allegro/bigcache/blob/master/queue/bytes_queue.go#L215) 

**Minimal, Complete, and Verifiable Example**
Not able to reproduce it 
```
	config := &bigcache.Config{
		LifeWindow:       2 * time.Hour,
		HardMaxCacheSize: 1024 * 2, 
		Shards:           1024,
	}
```
But was reading the code and thought this could be one of the reason

When asking a question about a problem caused by your code, you will get much better answers if you provide code we can use to reproduce the problem. That code should be...

* ...Minimal – Use as little code as possible that still produces the same problem
* ...Complete – Provide all parts needed to reproduce the problem
* ...Verifiable – Test the code you're about to provide to make sure it reproduces the problem

For more information on how to provide an MCVE, please see the [Stack Overflow documentation](https://stackoverflow.com/help/mcve).

**Environment:**
- Version (git sha or release):
- OS (e.g. from `/etc/os-release` or winver.exe): 
- go version:

mxplusb:
@paritosh-desai can you please provide a code example and the versions of bigcache you are using? Thanks.

paritosh-desai:
`https://github.com/allegro/bigcache/releases/tag/v2.1.4`

And not entirely sure how I can give a code example here
We are just calling the Set command.

PR #223, opened 2020-05-29
Reset full to false when pop from bytes queue

WideLee:
When byte queue is full, it will set `queue.full=true` but when evict and pop entry from queue, it does not set to false again. It causes bigcache cannot insert new entries once the queue is full.

This commit set `queue.full=false` when pop entry from queue, and add a test case which failed in current master https://github.com/allegro/bigcache/commit/ccdbc60304ad5ede288fea0a3d37ce4bae17b6db and passed in this commit.

Related to issue https://github.com/allegro/bigcache/issues/214
And it seems the same problem in Pull Request https://github.com/allegro/bigcache/pull/218

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -490,6 +490,33 @@ func TestCacheCapacity(t *testing.T) {
 	assertEqual(t, 40960, cache.Capacity())
 }
 
+func TestRemoveEntriesWhenShardIsFull(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cristaloleg (approved):
sgtm

siennathesane (approved):
lgtm, thanks for bug hunting!

PR #227, opened 2020-06-03
Fix typo in comment for Metadata struct

calebschoepp:
Just noticed this typo while reading through the docs.

Please let me know if there is anything else I need to do for this PR.


cristaloleg:
Thanks 🎉

Issue #226, opened 2020-06-03
panic: runtime error: index out of range [7] with length 1


jassentang:
Bigcache runs for about 1 day, crash！
How to solve this problem?
My bigcache configuration is:
```text
master_config := bigcache.DefaultConfig(300 * time.Second)
master_config.MaxEntriesInWindow = 10
master_config.HardMaxCacheSize = 2048
master_config.OnRemoveWithReason = func(key string, entry []byte, reason bigcache.RemoveReason) {
log.Info().Msgf("master bigcache OnRemove key[%s]entry[%s]reason[%d]Expired(1)NoSpace(2)Deleted(3)", key, entry, reason)
}
var initErr error
g_bigcache, initErr = bigcache.NewBigCache(master_config)
if initErr != nil {
log.Error().Err(initErr).Msgf("NewBigCache g_bigcache initErr!")
}
```
I used the Set, Get, Iterator(), SetNext() functions of bigcache.
The log of crash is as follows:
```text
2020/05/29 17:38:31 Allocated new queue in 84.16µs; Capacity: 2560000 6
2020/05/29 17:38:31 Allocated new queue in 130.649µs; Capacity: 2560000 6
2020/05/29 17:38:32 Allocated new queue in 197.943µs; Capacity: 4194304 9
2020/05/29 17:38:33 Allocated new queue in 130.092µs; Capacity: 4194304 9
panic: runtime error: index out of range [7] with length 1

goroutine 188 [running]:
panic(0x127da40, 0xc01a3417a0)
        /usr/local/go/src/runtime/panic.go:1064 +0x46d fp=0xc003264d88 sp=0xc003264cd0 pc=0x4461cd
runtime.goPanicIndex(0x7, 0x1)
        /usr/local/go/src/runtime/panic.go:88 +0xa3 fp=0xc003264dd0 sp=0xc003264d88 pc=0x443f73
encoding/binary.littleEndian.Uint64(...)
        /usr/local/go/src/encoding/binary/binary.go:77
github.com/allegro/bigcache/v2.readTimestampFromEntry(...)
        /root/go/pkg/mod/github.com/widelee/bigcache/v2@v2.2.1/encoding.go:43
github.com/allegro/bigcache/v2.(*cacheShard).onEvict(0xc0016777a0, 0xc00e23a860, 0x1, 0x3f37a0, 0x5ed0d923, 0xc003264e40, 0x0)
[...]

Issue #222, opened 2020-05-28
Iterator readEntry crashes


alexi:
**What is the issue you are having?**
Sometimes when using EntryInfoIterator to read all entries from the cache, the readEntry() call crashes with the following stack trace:

```
runtime error: makeslice: len out of range
  panic(0x1fabc00, 0x27ff350)
	/usr/local/go/src/runtime/panic.go:975 +0x3e3
  github.com/allegro/bigcache.readEntry(...)
  	/usr/local/Code/src/github.com/allegro/bigcache/encoding.go:36
  github.com/allegro/bigcache.(*EntryInfoIterator).Value(0xc10fd94a50, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
  	/usr/local/Code/src/github.com/allegro/bigcache/iterator.go:120 +0x305
```

We only recently updated from an old version of bigcache. This error seems to indicate the object being read is somehow larger than the maximum slice length, which should not be possible given we marshal the structs to bytes using `protobuf`'s `Marshal()` method:

```
func (m *CacheEntry) Marshal() (dAtA []byte, err error) {
	size := m.Size()
	dAtA = make([]byte, size)
	n, err := m.MarshalTo(dAtA)
	if err != nil {
		return nil, err
	}
	return dAtA[:n], nil
}
```
We are running on a 64bit system and we limit the size of all entries to ~1MB anyhow.

**What is BigCache doing that it shouldn't?**

It is not clear reading the code, but it seems related to how readEntry() determines length, perhaps some kind of overflow issue.

**Environment:**
- Version (git sha or release): v2.2.1
- OS (e.g. from `/etc/os-release` or winver.exe): ubuntu 16.04
- go version: 1.14.2

mxplusb:
Well that's exciting! Can you provide a quick reproduction of the bug so it's a bit easier to understand the full context? I think I understand where it's breaking, but a code example will help debug it a bit more.

alexi:
Hey @mxplusb - It's tricky to know how to reproduce it exactly as it is not something that occurs consistently so it seems that some unusual condition is met _sometimes_. Here is a mockup that represents roughly what we are doing:

```go
package main

import (
	"encoding/binary"
	"fmt"
	"math/rand"
	"time"

	"github.com/allegro/bigcache"
)

var cache *bigcache.BigCache

type Entry struct {
	id uint32
	v  []byte
}

func uint32Bytes(v uint32) []byte {
[...]

Issue #148, opened 2019-07-10
panic out of range in bytes queue


codyohl:
https://github.com/allegro/bigcache/blob/bbf64ae21fc5555f4e9752825ee9ffe13b1e5fa0/queue/bytes_queue.go#L222

It appears there is a bounds check before the blockSize, but then there is an assumption that adding block size is not out of bounds.

cristaloleg:
Hi, can you add a bit more details, please? What is the problem?

codyohlsen:
@cristaloleg, I can dig a little deeper into the bytes_queue to figure out what might be the right fix, but the underlying issue is that 
https://github.com/allegro/bigcache/blob/bbf64ae21fc5555f4e9752825ee9ffe13b1e5fa0/queue/bytes_queue.go#L223
results in a panic because` index+headerEntrySize+blockSize  >=len(q.array)` .

There is a bounds check on L218 here: https://github.com/allegro/bigcache/blob/bbf64ae21fc5555f4e9752825ee9ffe13b1e5fa0/queue/bytes_queue.go#L218

but there is no bounds check for adding the blockSize to that. Does that make sense?

cristaloleg:
Oh, finally I got it. I thought there is a problem on the line 222 as in the original message.

That's funny that there were no(???) panic before. I think somewhere in late 2017 I've checked this code and everything was fine. I might be wrong 😉

codyohlsen:
@cristaloleg it might be an uncommon edge case, or a misuse of the API somehow, but I did see this panic in logs using this code. If I have time, I can try to find a small reproducible test, it but it might be tough without enough knowledge about this code. I do experience this panic without changing this library, so some permutation of the usage of this API will cause the panic

olivierperes:
@cristaloleg I obtain this panic while using an iterator to check all entries whose name match a given pattern and deleting them, while other processes are adding entries. It is also possible that two of these iterators work concurrently, but they cannot try to delete the same entries. Is this a supported pattern?

janisz:
@olivierperes Can you provide code to demonstrate this behavior?

olivierperes:
This is the function that deletes entries. Unfortunately I cannot provide a unit test. My program can run for about 12 hours before the crash occurs.
```go
// cleanCache deletes from the cache all elements whose name has the given prefix.
func cleanCache(prefix string) error {
	iterator := cache.Iterator()
	for iterator.SetNext() {
		entry, err := iterator.Value()
		if err != nil {
[...]

PR #232, opened 2020-06-25
fix: panic when update and iterate simultaneously

WideLee:
1. Fix panic when update and iteration simultaneously. Related to https://github.com/allegro/bigcache/issues/222.
2. Add panic recover in cleanUp to prevent the program exited. Related to https://github.com/allegro/bigcache/issues/226, https://github.com/allegro/bigcache/issues/148, just protect the main program not exited.
3. Byte queue set full is false after allocated addition memory. Also added a test case reproduce this problem.

### DIFF
--- a/shard.go
+++ b/shard.go
@@ -264,20 +264,25 @@ func (s *cacheShard) cleanUp(currentTimestamp uint64) {
 	s.lock.Unlock()
 }
 
-func (s *cacheShard) getEntry(index int) ([]byte, error) {
+func (s *cacheShard) getEntry(hashedKey uint64) ([]byte, error) {
 	s.lock.RLock()
-	entry, err := s.entries.Get(index)
+
+	entry, err := s.getWrappedEntry(hashedKey)
+	// copy entry
+	newEntry := make([]byte, len(entry))
+	copy(newEntry, entry)
+
 	s.lock.RUnlock()
 
-	return entry, err
+	return newEntry, err
 }
 
-func (s *cacheShard) copyKeys() (keys []uint32, next int) {
+func (s *cacheShard) copyHashedKeys() (keys []uint64, next int) {
 	s.lock.RLock()
-	keys = make([]uint32, len(s.hashmap))
+	keys = make([]uint64, len(s.hashmap))
 
-	for _, index := range s.hashmap {
-		keys[next] = index
+	for key := range s.hashmap {
+		keys[next] = key
 		next++
 	}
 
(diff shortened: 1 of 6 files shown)


shard.go
```diff
@@ -253,6 +254,15 @@ func (s *cacheShard) onEvict(oldestEntry []byte, currentTimestamp uint64, evict
 }
 
 func (s *cacheShard) cleanUp(currentTimestamp uint64) {
+	defer func() {
+		// panic recover
```
janisz, line 258: Why do we need this?

WideLee: In https://github.com/allegro/bigcache/issues/226 and https://github.com/allegro/bigcache/issues/148, It panic in `cleanUp`.

If CleanWindows is not zero, bigcache starts a goroutine, if cleanUp panic, the whole program will be exited. https://github.com/allegro/bigcache/blob/bbbffd30d25b538b4c03be6649b1dffbacca9596/bigcache.go#L85-L99

This is a temporary solution to prevent crash of whole program, actually, find out the reason why panic is important.

WideLee:
> I think it looks good to me, it could just use adjustments based off @janisz's comments.

Do you mean that I should rollback the commit about panic recover in `CleanUp`?

WideLee:
@janisz @mxplusb  I just reverted the panic recover commit, please review it again 😄

alexi:
Is there any progress on solving this underlying issue?  We are still getting these crashes 🤦

mxplusb:
@alexi you're still having this same issue?

alexi:
@mxplusb now getting this error: runtime error: slice bounds out of range [:46697172] with capacity 585000

mxplusb:
@alexi can you please verify you are using this version or latest?

alexi:
Yes, using commit 7bf29c05ed83207873911f27b20170baedd4df0a (most recent).

The issues section has multiple examples of this bug: #226, and #148 has an open pull request fixing a uint32 overflow which looks like the likely culprit.

mxplusb:
@alexi yeah, I think #148 could fix this.

alexi:
@mxplusb Any idea when it might be merged?

fpessolano:
Any news, we have now the same issues on production servers (which we let crash and restart at the moment)

alexi:
Hey @mxplusb any update here?  We are in the same boat as @fpessolano.  If it's possible to deploy it as a smaller patch that would be very valuable.

mxplusb:
We're prepping for v3 in the `versions/v3` branch, which should include the fix for this, but no current ETA on the release. I might be able to spend some time on it this month.

PR #237, opened 2020-08-26
Fix in initial capacity allocation

ac-rappi:
Currently there is a mix between hard limit in memory and amount of entries used for initial capacity allocation:

initialShardSize() does a "min" on: bytes (From max shard size calculation) and amount of entries.

This leads to breaching the HardMaxCacheSize contract

### DIFF
--- a/config.go
+++ b/config.go
@@ -68,11 +68,11 @@ func DefaultConfig(eviction time.Duration) Config {
 
 // initialShardSize computes initial shard size
 func (c Config) initialShardSize() int {
-	return max(min(c.MaxEntriesInWindow/c.Shards, c.maximumShardSize()), minimumEntriesInShard)
+	return max(c.MaxEntriesInWindow/c.Shards, minimumEntriesInShard)
 }
 
-// maximumShardSize computes maximum shard size
-func (c Config) maximumShardSize() int {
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -68,11 +68,11 @@ func DefaultConfig(eviction time.Duration) Config {
 
 // initialShardSize computes initial shard size
 func (c Config) initialShardSize() int {
-	return max(min(c.MaxEntriesInWindow/c.Shards, c.maximumShardSize()), minimumEntriesInShard)
+	return max(c.MaxEntriesInWindow/c.Shards, minimumEntriesInShard)
 }
 
-// maximumShardSize computes maximum shard size
-func (c Config) maximumShardSize() int {
+// maximumShardSizeInBytes computes maximum shard size
```
zeitos, line 74: nit: "... shard size in bytes"

janisz (approved):

cristaloleg (approved):

ac-rappi:
Thanks @cristaloleg ! Do you think a tag including this fix is coming soon?

cristaloleg:
It's created automatically, see: https://github.com/allegro/bigcache/releases/tag/v2.2.4 :)

PR #242, opened 2020-09-12
export clock method to allow assignable in other packages

### LINKED ISSUE #241: support mocking clock in other projects
I'm doing unit testing, and I cannot find a way to mock `Bigcache#clock`.

* the `Bigcache#clock` attribute is unexported
* the `newBigcache(...)` is unexported
* the methods in `clock interface` is unexported, and the [reflect#implements](https://github.com/golang/go/blob/dfdc3880b01d46d1d8125ab9eea0606b2fa5b819/src/reflect/type.go#L1501) rejects due to difference packages. 

I proposal that we export `clock#Epoch`, and it seems have no side effects.

Any suggestions? I'm glad to contribute a patch for this.

flisky:
For example:

```
cache := bigcache.NewBigCache(bigcache.DefaultConfig(time.Minute))
clock := &mockedClock{}
rf := reflect.ValueOf(cache).Elem().FieldByName("clock")
reflect.NewAt(rf.Type(), unsafe.Pointer(rf.UnsafeAddr())).Elem().Set(reflect.ValueOf(clock))
```

Without this change, it panics with `value of type *cache_test.mockedClock is not assignable to type bigcache.clock`

And it goes to `*cache_test.mockedClock cannot implement bigcache.Clock as it has a non-exported method and is defined in a difference package` if we only export clock interface.

Fix #241

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -389,7 +389,7 @@ func TestOnRemoveFilterExpired(t *testing.T) {
 
 	cache.Set("key", []byte("value"))
 	clock.set(5)
-	cache.cleanUp(uint64(clock.epoch()))
+	cache.cleanUp(uint64(clock.Epoch()))
 
 	err = cache.Delete("key")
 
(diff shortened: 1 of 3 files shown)


janisz (approved):

mxplusb (approved):
LGTM.

PR #236, opened 2020-08-21
Use uint64 intead of uint32

janisz:
There are posibility we run into a problem of int32 overflow.
To prevent this let's use uint64 everywhere.

https://github.com/allegro/bigcache/blob/21e5ca5c3d539f94e8dc563350acd97c5400154f/shard.go#L138

Fixes: https://github.com/allegro/bigcache/issues/148

Issue #235, opened 2020-08-19
bigcache panic for slice 


LIUHUANUCAS:
**What is the issue you are having?**
```
panic: runtime error: slice bounds out of range

goroutine 229 [running]:
github.com/allegro/bigcache/queue.(*BytesQueue).peek(0xc4201e25a8, 0x36, 0xc425e9e480, 0xc420055900, 0x44e395, 0xc42a35d110, 0x1b39d171, 0x198977d7a03ddf50)
        ****/src/github.com/allegro/bigcache/queue/bytes_queue.go:195 +0x22e
github.com/allegro/bigcache/queue.(*BytesQueue).Get(0xc4201e25a8, 0x36, 0xc09f58197aa79848, 0xc4201be20c, 0xc427c818a8, 0xc42a35d160, 0x50054c)
        ****src/github.com/allegro/bigcache/queue/bytes_queue.go:161 +0x35
github.com/allegro/bigcache.(*cacheShard).get(0xc4201e25a0, 0xc42ea01c70, 0xb, 0xc09f58197aa79848, 0xc42a35d268, 0x4f6d86, 0x5f3cdd9a, 0xce7cfca, 0x432f257b0d)
        ****/src/github.com/allegro/bigcache/shard.go:38 +0xa7
github.com/allegro/bigcache.(*BigCache).Get(0xc4202b2000, 0xc42ea01c70, 0xb, 0x9, 0xbd2a20, 0xe4f290, 0xc42b9b0900, 0xc425e13650)
        ****/src/github.com/allegro/bigcache/bigcache.go:93 +0x8a
main.getDocidlist(0xc42ea01c70, 0xb, 0x0, 0x0, 0x0, 0x0, 0x0)
```

**What is BigCache doing that it shouldn't?**

**Minimal, Complete, and Verifiable Example**

When asking a question about a problem caused by your code, you will get much better answers if you provide code we can use to reproduce the problem. That code should be...

* ...Minimal – Use as little code as possible that still produces the same problem
* ...Complete – Provide all parts needed to reproduce the problem
* ...Verifiable – Test the code you're about to provide to make sure it reproduces the problem

For more information on how to provide an MCVE, please see the [Stack Overflow documentation](https://stackoverflow.com/help/mcve).

**Environment:**
- go env 
- bigcache version 2 years ago
```
// go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
[...]

Issue #234, opened 2020-07-28
v2.2.3 panic


work-chausat:
panic: runtime error: index out of range [7] with length 7

goroutine 491 [running]:
encoding/binary.littleEndian.Uint64(...)
        /export/App/go/src/encoding/binary/binary.go:77
github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2.readTimestampFromEntry(...)
        /export/App/gopath/src/github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2/encoding.go:43
github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2.(*cacheShard).onEvict(0xc12892c360, 0xc1b072c275, 0x7, 0x14fb, 0x5f1fc08d, 0xc136bf8ad0, 0x0)
        /export/App/gopath/src/github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2/shard.go:247 +0x7d
github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2.(*cacheShard).set(0xc12892c360, 0xc0669891f0, 0xd, 0x97320e0161446698, 0xc0669891f0, 0xd, 0xc136bf8e30, 0xf273c0, 0xc000199ce0)
        /export/App/gopath/src/github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2/shard.go:131 +0x30f
github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2.(*BigCache).Set(0xc11cab3860, 0xc0669891f0, 0xd, 0xc0669891f0, 0xd, 0xc136bf8e30, 0xf273c0, 0xc000199ce0)
        /export/App/gopath/src/github.com/baudtime/baudtime/vendor/github.com/allegro/bigcache/v2/bigcache.go:133 +0xac

janisz:
duplicates https://github.com/allegro/bigcache/issues/226

PaulLiang1:
Could be reproduced with following
```
package main

import (
	"fmt"
	"math/rand"
	"runtime"
	"strconv"
	"time"

	. "github.com/allegro/bigcache/v2"
	"github.com/pkg/profile"
)

const (
	entries = 3000
	repeats = 1000
)

var val_big = append(make([]byte, 100 * 1024), 1)
var val_med = append(make([]byte, 1024), 1)
var val_small = append(make([]byte, 2), 1)

func get_value() []byte {
[...]

PR #207, opened 2020-03-12
Optimize bytes_queue

neal-zhu:
Hi there, I try to optimize bytes_queue in two ways:
1. The length is encoded in varint format(the assertion on capacity will fail so I commented them out).
2. When we insert an entry before the head of the queue, and the size of this entry exactly equals to unsued bytes, there is no need to reserve bytes. 

And below are the outputs of test on my Mac:
```
# Mine
2020/03/12 14:49:28 Allocated new queue in 130ns; Capacity: 88
2020/03/12 14:49:28 Allocated new queue in 211.918µs; Capacity: 585000
2020/03/12 14:49:28 Allocated new queue in 394.027µs; Capacity: 1170000
PASS
```

```
# Yours
2020/03/12 14:06:56 Allocated new queue in 156ns; Capacity: 94
2020/03/12 14:06:56 Allocated new queue in 719.26µs; Capacity: 585000
2020/03/12 14:06:56 Allocated new queue in 566.888µs; Capacity: 1170000
PASS
```

PS: It can save some bytes and avoid re-allocation of queue at some time

PR #251, opened 2020-10-30
Fix a bug that obscures the byte queue when it is full and is extended

Fabianexe:
Hi,
On searching for the reason for the panics in the current version where versions pre 2.2 works well, I think I have found the cause. 
With PR #207 it gets possible that the byte queen is full and `q.tail == q.head`. However, in the `allocateAdditionalMemory` method, the condition for resetting head and tail is `q.tail < q.head`. Thus the tail and head are not changed, and the byte queen is obscured after it. An example that leads to panic in the old logic is shown in the test `TestPushEntryAfterAllocateAdditionMemoryInFull`.

This is a different reason for panics then #236, so I am not sure which of the reported panic ( #226 #234 #235 ) is caused by this bug.

### DIFF
--- a/queue/bytes_queue.go
+++ b/queue/bytes_queue.go
@@ -8,7 +8,7 @@ import (
 
 const (
 	// Number of bytes to encode 0 in uvarint format
-	minimumHeaderSize = 1
+	minimumHeaderSize = 17 // 1 byte blobsize + timestampSizeInBytes + hashSizeInBytes
 	// Bytes before left margin are not used. Zero index means element does not exist in queue, useful while reading slice from index
 	leftMarginIndex = 1
 )
--- a/queue/bytes_queue.go
+++ b/queue/bytes_queue.go
@@ -118,10 +118,13 @@ func (q *BytesQueue) allocateAdditionalMemory(minimum int) {
 	if leftMarginIndex != q.rightMargin {
 		copy(q.array, oldArray[:q.rightMargin])
 
-		if q.tail < q.head {
-			headerEntrySize := getUvarintSize(uint32(q.head - q.tail))
-			emptyBlobLen := q.head - q.tail - headerEntrySize
-			q.push(make([]byte, emptyBlobLen), emptyBlobLen)
+		if q.tail <= q.head {
+			if q.tail != q.head {
+				headerEntrySize := getUvarintSize(uint32(q.head - q.tail))
+				emptyBlobLen := q.head - q.tail - headerEntrySize
+				q.push(make([]byte, emptyBlobLen), emptyBlobLen)
+			}
+
 			q.head = leftMarginIndex
 			q.tail = q.rightMargin
 		}
--- a/queue/bytes_queue_test.go
+++ b/queue/bytes_queue_test.go
@@ -389,8 +389,6 @@ func TestPushEntryAfterAllocateAdditionMemory(t *testing.T) {
 	queue.Push([]byte("aaa"))
 	queue.Push([]byte("bb"))
 	queue.Pop()
-	queue.Push([]byte("c"))
-	queue.Push([]byte("d"))
 
 	// allocate more memory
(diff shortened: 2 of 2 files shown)


janisz:
LGTM

I've tested it with test included in #234 but it does not solve that issues

<details>

```go
package bigcache

import (
	"math/rand"
	"runtime"
	"strconv"
	"testing"
	"time"
)

const (
	entries = 3000
	repeats = 1000
)

var valBig = append(make([]byte, 100*1024), 1)
var valMed = append(make([]byte, 1024), 1)
var valSmall = append(make([]byte, 2), 1)

func getValue() []byte {
	x := rand.Float64()
	if x < 0.7 {
		return valSmall
	} else if x < 0.9 {
		return valMed
	}
	return valBig
}

// https://github.com/allegro/bigcache/issues/234#issuecomment-673895517
func Test_issue_234(t *testing.T) {
	t.Log("Number of entries: ", entries)
	printAllocs(t)
	rand.Seed(1)

	config := Config{
		Shards:             128,
		LifeWindow:         time.Hour,
		CleanWindow:        time.Second,
		MaxEntriesInWindow: entries,
		MaxEntrySize:       1024,
		Verbose:            true,
		HardMaxCacheSize:   128,
		OnRemoveWithReason: func(key string, entry []byte, reason RemoveReason) {
			t.Log("Evicted:", len(entry), " reason: ", reason)
		},
	}

	bigcache, err := NewBigCache(config)
	if err != nil {
		panic(err)
	}
	for i := 0; i < repeats; i++ {
		printAllocs(t)
		for j := 0; j < entries; j++ {
			key := strconv.FormatInt(int64(j), 10)
			err := bigcache.Set(key, getValue())
			if err != nil {
				t.Fatal(err)
			}
		}
	}
}

func printAllocs(t *testing.T) {
	var m runtime.MemStats
	runtime.ReadMemStats(&m)
	t.Logf("Alloc: %6d MB \n", m.Alloc/1e6)
}
```
</details>

janisz (dismissed):

Fabianexe:
@janisz 
I have relooked at the test that you suggested. 
This is a related but a little different bug.
By reallocating the byte queen, an artificial entry is added to close the gap between the head and tail. This artificial entry could have no content. This is fine for the byte queen, but by calling Pop this entry is published to the shard, and this shard then expects content.
I have pushed a fix for this bug as well. Simple set the minimum size for such entry to 17 (size+timestamp+hash).

A more minimalistic test for the bug from #234 is:

```
package bigcache

import (
	"testing"
	"time"
)

// https://github.com/allegro/bigcache/issues/234#issuecomment-673895517
func Test_issue_234(t *testing.T) {

	config := Config{
		Shards:             1,
		LifeWindow:         time.Hour,
		MaxEntriesInWindow: 900,
		MaxEntrySize:       1024,
		Verbose:            true,
		HardMaxCacheSize:   1,
		OnRemoveWithReason: func(key string, entry []byte, reason RemoveReason) {
			t.Log("Evicted:", len(entry), " reason: ", reason)
		},
	}

	bigcache, _ := NewBigCache(config)
	bigcache.Set("1", make([]byte, 400*1024))
	bigcache.Set("2", make([]byte, 400*1024))
	bigcache.Set("3", make([]byte, 400*1024-1))
	bigcache.Set("4", make([]byte, 100*1024))
	bigcache.Set("5", make([]byte, 400*1024))
}
```

janisz (approved):

janisz:
@mxplusb @cristaloleg Can you look at this?

mxplusb (approved):
LGTM! Thanks!

Fabianexe:
@mxplusb 
I would love to see a new release with this so that bigcache can simple update

mxplusb:
@Fabianexe I released v2.2.5 just now, which includes this fix, so you should be good to go!

janisz:
Refs: #253

PR #265, opened 2021-02-10
Skip hashing

SlashGordon:
I need a way to provide my own hash value without string to uint64 conversion.
I think the simplest solution is to provide setter and getter with the ability to skip hashing.

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -113,8 +113,7 @@ func (c *BigCache) Close() error {
 // no entry exists for the given key.
 func (c *BigCache) Get(key string) ([]byte, error) {
 	hashedKey := c.hash.Sum64(key)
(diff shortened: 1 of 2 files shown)


cristaloleg (changes requested):
See linter fails

SlashGordon:
I have fixed the linter errors :)

cristaloleg (dismissed):

cristaloleg:
@janisz @mxplusb if ok - let's merge

janisz:
I'm not sure about it. If you need to get set without key why you propose to function that requires key and hashed key?
Can't you just use custom hasher?

If this is about to land I think we should use `GetByHash` in `Get` just to simplify code and reduce needed tests

Issue #253, opened 2020-11-13
when invoke allocateAdditionalMemory() , if the free space is 128 bytes, bug occurs


KaixuanYu:
The `allocateAdditionalMemory` method in `bigcache/queue/bytes_queue.go` file has the following code:
```
		if q.tail <= q.head {
			if q.tail != q.head {
				headerEntrySize := getUvarintSize(uint32(q.head - q.tail))
				emptyBlobLen := q.head - q.tail - headerEntrySize
				q.push(make([]byte, emptyBlobLen), emptyBlobLen)
			}

			q.head = leftMarginIndex
			q.tail = q.rightMargin
		}
```
When head - tail == 128, the q.push() function actually only stores 127 bytes, one less byte.
A bug occurs when you call the Pop function。
See the following unit test for details：
```
func TestAllocateAdditionalSpaceForInsufficientFreeFragmentedSpaceWhereTailIsBeforeHead(t *testing.T) {
	t.Parallel()

	// given
	queue := NewBytesQueue(200, 0, false)

	// when 
	queue.Push(blob('a', 30)) //  header + entry + left margin = 32 bytes
	queue.Push(blob('b', 127)) // 32 + 127 + 1 = 160 bytes
	queue.Push(blob('c', 20)) //  160 + 20 + 1 = 181
	queue.Pop()               // space freed at the beginning
	queue.Pop()               //
	queue.Push(blob('d', 30)) // 31 bytes used at the beginning, tail pointer is before head pointer, now free space is 128 bytes
	queue.Push(blob('e', 160)) // invoke allocateAdditionalMemory but fill 127 bytes free space (It should be 128 bytes, but 127 are filled, leaving one byte unfilled)

	// then
	//assertEqual(t, 400, queue.Capacity())
	//assertEqual(t, blob('d', 30), pop(queue))
	//assertEqual(t, blob(0, 127), pop(queue))    //error
	//assertEqual(t, blob('c', 20), pop(queue))   //The data is not expected
	//assertEqual(t, blob('e', 160), pop(queue))
}
```

janisz:
Refs #251

KaixuanYu:
If head-tail == 129, I think there is no way to deal with it. Because no length is 129 byte after being encoded with varint.

janisz:
So it looks like the only way to fix this is to revert #207 and not use Varint

panmari:
[...]

PR #266, opened 2021-02-11
The size encoded in a entry header contains the header it self.

Fabianexe:
As #253 has pointed out, it is impossible to create an entry of arbitrary size in the byte queue. Basically, the problem is that at the points where the header gets larger, one number is skipped. This can lead to errors in the new allocation of a byte queue.

To solve this, the header size is included in the number that is encoded. This allows creating entries with arbitrary dimensions. For example, we have 127 bytes (1 header, 126 data), 128bytes (2 header, 126 data), and 129 bytes(2 header, 127 data).

As a benefit, we can reduce some calculations, and the code gets more straightforward. 
The only downside is that the maximum entry size reduces from `4294967296` to `4294967291` bytes. In my belief, nothing too dramatic that gets in v3 with uint64, even a more minor problem.

### DIFF
--- a/queue/bytes_queue.go
+++ b/queue/bytes_queue.go
@@ -38,19 +38,23 @@ type queueError struct {
 	message string
 }
 
-// getUvarintSize returns the number of bytes to encode x in uvarint format
-func getUvarintSize(x uint32) int {
-	if x < 128 {
-		return 1
-	} else if x < 16384 {
-		return 2
-	} else if x < 2097152 {
-		return 3
-	} else if x < 268435456 {
-		return 4
-	} else {
-		return 5
+// getNeededSize returns the number of bytes an entry of length need in the queue
+func getNeededSize(length int) int {
+	var header int
+	switch {
+	case length < 127: // 1<<7-1
+		header = 1
+	case length < 16382: // 1<<14-2
+		header = 2
+	case length < 2097149: // 1<<21 -3
+		header = 3
+	case length < 268435452: // 1<<28 -4
+		header = 4
+	default:
+		header = 5
 	}
+
+	return length + header
 }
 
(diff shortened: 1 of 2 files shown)


Fabianexe:
@KaixuanYu @janisz

user3415653:
Any news here? Sometimes we get this error, so a fix would be nice.

janisz (approved):

janisz:
@Fabianexe Could you rebase?
@cristaloleg Could you take a look?

ekkoxx (commented):
fix

panmari:
Anything else needed here? I'd love to see this merged.

cristaloleg (approved):

panmari:
On HEAD, I'm still seeing the error `panic: runtime error: index out of range [7] with length 0` with the trace 

```
encoding/binary.littleEndian.Uint64(...)
encoding/binary/binary.go:77
bigcache/bigcache.readTimestampFromEntry(...)
bigcache/encoding.go:58
bigcache/bigcache.(*cacheShard).onEvict(0xc008d6e008?, {0xc27fe50e69?, 0x1c6b0668dbacaecc?, 0x0?}, 0x5593c1f7d065?, 0xc031bc5888?)
bigcache/shard.go:271 +0x65
bigcache/bigcache.(*cacheShard).set(0xc008d6e000, {0xc0510a08a0, 0x25}, 0x1c6b0668dbacaecc, {0x0, 0x0, 0x25?})
bigcache/shard.go:134 +0x197
bigcache/bigcache.(*BigCache).Set(0xc002a149c0, {0xc0510a08a0, 0x25}, {0x0, 0x0, 0x0})
```

What's blocking here? Do you want me to do some more evaluation before merging?

PR #268, opened 2021-02-27
Update readme to version 3

mhrlife:
We had some memory leaks in production. after further investigations, we found out that we were using big cache v1 (which interestingly HardMaxCacheSize is not working properly). I'm sure if we don't change it to version 2 it will cause some confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Requires Go 1.12 or newer.
 ### Simple initialization
 
 ```go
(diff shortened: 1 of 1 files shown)


janisz:
We already have V3 
Could you update to the latest version?

cristaloleg (approved):

janisz (approved):

cristaloleg:
Thank you

PR #287, opened 2021-08-15
implement to save slices

elvin-zheng:


janisz:
@elvin-zheng Could you add description?
@cristaloleg Could you run the CI?

bigcache.go
```diff
@@ -238,3 +240,67 @@ func (c *BigCache) providedOnRemoveWithMetadata(wrappedEntry []byte, reason Remo
...
+		}
+		bs,_:=json.Marshal(u)
+		userBs = append(userBs,bs)
+	}
+	cache.PutLists("users",userBs...)
+
+	entries,err:=cache.GetLists("users")
+*/
+
+func (c *BigCache) PutLists(key string, entries ...[]byte) (err error) {
```
janisz, line 265: Why do we need that?

elvin-zheng: After appending, the bytes without boundary is obtained

Issue #293, opened 2021-10-27
Remove `maxShardSize` from BigCache


janisz:
It looks like `BigCache.maxShardSize` is only set but never queried so it can be safely deleted.

https://github.com/allegro/bigcache/blob/c2b72fe951a1ff765a0e58184738476f2fd71bb1/bigcache.go#L22
https://github.com/allegro/bigcache/blob/c2b72fe951a1ff765a0e58184738476f2fd71bb1/bigcache.go#L66

PR #294, opened 2021-10-28
Remove `maxShardSize` from BigCache

penthaapatel:
Fix: #293


janisz:
@penthaapatel Thanks 🚀

PR #299, opened 2021-11-14
Adding config validation when creating new cache.

panmari:
Otherwise this causes more cryptic failures in later code. Best to catch this on initialization.


janisz (commented):
Can you add a test for that change?

panmari:
Added test coverage.

janisz (commented):
How about using table test for that?

panmari:
Sure thing, switched to a table test.

janisz (approved):

janisz:
We need to also fix comments https://github.com/allegro/bigcache/blob/master/config.go#L12

PR #314, opened 2022-04-15
Fix subtractive overflow

cancanyang:
There is a possibility that currentTimestamp is smaller than oldestTimestamp, creating an overflow and causing normal data to be cleaned up

### DIFF
--- a/shard.go
+++ b/shard.go
@@ -269,7 +269,10 @@ func (s *cacheShard) del(hashedKey uint64) error {
 
 func (s *cacheShard) onEvict(oldestEntry []byte, currentTimestamp uint64, evict func(reason RemoveReason) error) bool {
 	oldestTimestamp := readTimestampFromEntry(oldestEntry)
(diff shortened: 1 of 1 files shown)


shard.go
```diff
@@ -269,7 +269,10 @@ func (s *cacheShard) del(hashedKey uint64) error {
 
 func (s *cacheShard) onEvict(oldestEntry []byte, currentTimestamp uint64, evict func(reason RemoveReason) error) bool {
 	oldestTimestamp := readTimestampFromEntry(oldestEntry)
-	if currentTimestamp-oldestTimestamp > s.lifeWindow {
+	keyHash := readHashFromEntry(oldestEntry)
+	if keyHash == 0 || (currentTimestamp >= oldestTimestamp &&
+		currentTimestamp-oldestTimestamp > s.lifeWindow) {
+
```
janisz, line 275: Could you add a test for it?

janisz:
@cancanyang  ping

PR #317, opened 2022-05-07
Fix bug to remove non-expired data

1xxz188:
Sometimes an error occurs to remove non-expired data when "currentTimestamp" is less than "oldestTimestamp"


cristaloleg (approved):
sgtm

janisz (approved):

PR #325, opened 2022-07-29
No need to del elem when config.CleanWindow>0 at method:Set

zhangyaping-CN:


cristaloleg (commented):
SGTM, maybe a test?

shard.go
```diff
@@ -129,8 +130,10 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
 		}
 	}
 
-	if oldestEntry, err := s.entries.Peek(); err == nil {
-		s.onEvict(oldestEntry, currentTimestamp, s.removeOldestEntry)
+	if !s.cleanEnabled {
```
janisz, line 133: Could you add a test for this?

janisz (changes requested):

shard.go
zhangyaping-CN: Thank you for your reply, I added testing and then added `if !s.cleanEnabled` when Append

cristaloleg (commented):
SGTM

janisz (approved):

PR #326, opened 2022-08-10
Add method SetCb

1xxz188:
The purpose is to support some atomic operations


cristaloleg:
Can you clarify the use case? Why it cannot be just:

```
err := cache.Set("key", []byte("value"))
callback()
```

?

1xxz188:
> 
in order to verify that it is not affected when cb was nil.

1xxz188:
Long story, wait until I have time to reply.

Issue #336, opened 2022-10-04
Shutting down cleanup routine with ctx


alok87:
**What is the issue you are having?**
No issue, saw the NewBigCache() creates a go routine but does not take the context from the users and does not do a graceful shutdown of that go routine on context cancellation.
```go
// NewBigCache initialize new instance of BigCache
func NewBigCache(config Config) (*BigCache, error) {
	return newBigCache(config, &systemClock{})
}
```

**Expected**
```go
// NewBigCache initialize new instance of BigCache
func NewBigCache(ctx context.Context, config Config) (*BigCache, error) {
	return newBigCache(config, &systemClock{})
}

go func() {
			ticker := time.NewTicker(config.CleanWindow)
			defer ticker.Stop()
			for {
				select {
                                 case <-ctx.Done():
                                         fmt.Println("ctx done, shutting down")
                                         return
				case t := <-ticker.C:
					cache.cleanUp(uint64(t.Unix()))
				case <-cache.close:
					return
				}
			}
		}()
```

**Environment:**
- Version (git sha or release): v3.2.0
- OS (e.g. from `/etc/os-release` or winver.exe): macOS 
- go version: 1.19

janisz:
@alok87 Good point. When bigcache was created there were no context in std lib. Now we should support that to prevent gorutine leakage. 

@alok87 Would you like to contribute a PR for this?

alok87:
Sure, will give a PR

PR #337, opened 2022-10-04
Cleanly shutdown cleanup go-routine

alok87:
Cleanly shutdown go routine. Big cache was made when go did not have ctx support

Fixes https://github.com/allegro/bigcache/issues/336

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -1,6 +1,7 @@
 package bigcache
 
 import (
+	"context"
 	"fmt"
 	"time"
 )
@@ -41,11 +42,11 @@ const (
 )
 
 // NewBigCache initialize new instance of BigCache
-func NewBigCache(config Config) (*BigCache, error) {
-	return newBigCache(config, &systemClock{})
+func NewBigCache(ctx context.Context, config Config) (*BigCache, error) {
+	return newBigCache(ctx, config, &systemClock{})
 }
 
-func newBigCache(config Config, clock clock) (*BigCache, error) {
+func newBigCache(ctx context.Context, config Config, clock clock) (*BigCache, error) {
 	if !isPowerOfTwo(config.Shards) {
 		return nil, fmt.Errorf("Shards number must be power of two")
 	}
@@ -94,6 +95,9 @@ func newBigCache(config Config, clock clock) (*BigCache, error) {
 			defer ticker.Stop()
 			for {
 				select {
+				case <-ctx.Done():
+					fmt.Println("ctx done, shutting down bigcache cleanup routine")
+					return
 				case t := <-ticker.C:
 					cache.cleanUp(uint64(t.Unix()))
 				case <-cache.close:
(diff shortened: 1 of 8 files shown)


bigcache.go
```diff
@@ -41,11 +42,11 @@ const (
 )
 
 // NewBigCache initialize new instance of BigCache
-func NewBigCache(config Config) (*BigCache, error) {
```
janisz, line 44: I think we should keep the old constructor for backward compatibility but mark it deprecated

bigcache_bench_test.go
```diff
@@ -16,7 +17,7 @@ func BenchmarkWriteToCacheWith1Shard(b *testing.B) {
 
 func BenchmarkWriteToLimitedCacheWithSmallInitSizeAnd1Shard(b *testing.B) {
 	m := blob('a', 1024)
-	cache, _ := NewBigCache(Config{
+	cache, _ := NewBigCache(context.TODO(), Config{
```
janisz, line 20: I think in tests its fine to use `context.Background()`

bigcache.go
alok87: name suggestion for the new function `NewBigCacheWithContext()`? 

and what should we pass in the old function TODO() or Background() is fine?

alok87: or just `New()` as the package name is bigcache

`bigcache.New()`

janisz: Let's go with `WithContext` suffix as this is how std lib added context https://godocs.io/net/http#NewRequestWithContext

alok87: Both background and todo returns the same empty context
```
var (
	background = new(emptyCtx)
	todo       = new(emptyCtx)
)
```
https://go.dev/src/context/context.go

cristaloleg: It's not about implementation, it's about readability 😉

bigcache.go
```diff
@@ -40,12 +41,19 @@ const (
...
 )
 
+// New initialize new instance of BigCache
+func New(ctx context.Context, config Config) (*BigCache, error) {
+	return newBigCache(ctx, config, &systemClock{})
+}
+
 // NewBigCache initialize new instance of BigCache
+// deprecated, please use New(ctx, config) instead, New takes in context and can gracefully
+// shutdown with context cancellations
```
janisz, line 51:
```suggestion
// NewBigCache initialize new instance of BigCache
//
// Deprecated: NewBigCache is deprecated, please use New(ctx, config) instead, New takes in context and can gracefully
// shutdown with context cancellations
```
https://github.com/golang/go/wiki/Deprecated

janisz (changes requested):

bigcache.go
alok87: Fixed

janisz (approved):

bigcache.go
alok87: Saw the comment in `WithContext` now, went ahead with `New()`, looks more redable to me. Let me know if it requires a change.

Std lib follows both, see this https://pkg.go.dev/errors#New 
I think it depends on case by case basis. Also for `http.New()` wont really work as it is a multiple object inside http.

Issue #341, opened 2022-10-12
Update actions with dependabot


janisz:
Configure @dependabot to update actions.
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
Refs: #340

janisz:
Fixed by #342

PR #342, opened 2022-10-12
Configure Dependabot for GH Actions (#341)

madhead:


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+
+updates:
```
cristaloleg, line 3: Feel free to add `Go` too, just in case 😉

cristaloleg (approved):

.github/dependabot.yml
madhead: Done

janisz (approved):

PR #372, opened 2023-08-21
Use pool instead of entryBuffer

valiner:
Use sync.pool to get higher performance


cristaloleg:
Hey, thanks for the feature. Can you re-run benchmarks?

valiner:
> Hey, thanks for the feature. Can you re-run benchmarks?

I reconfirmed it. It's not a good idea.

PR #374, opened 2023-09-11
Stop printing message when doing a context shutdown

rubenv:
Uses the configured logger instead.

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -3,7 +3,6 @@ package bigcache
 import (
 	"context"
 	"errors"
-	"fmt"
 	"time"
 )
 
@@ -101,12 +100,13 @@ func newBigCache(ctx context.Context, config Config, clock clock) (*BigCache, er
 
 	if config.CleanWindow > 0 {
 		go func() {
+			logger := newLogger(config.Logger)
 			ticker := time.NewTicker(config.CleanWindow)
 			defer ticker.Stop()
 			for {
 				select {
 				case <-ctx.Done():
-					fmt.Println("ctx done, shutting down bigcache cleanup routine")
+					logger.Printf("ctx done, shutting down bigcache cleanup routine")
 					return
 				case t := <-ticker.C:
 					cache.cleanUp(uint64(t.Unix()))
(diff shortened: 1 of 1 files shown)


cristaloleg:
Good catch, thanks!

cristaloleg (dismissed):

rubenv:
I honestly have no idea why it would cause that.

bigcache.go
```diff
@@ -101,12 +100,13 @@ func newBigCache(ctx context.Context, config Config, clock clock) (*BigCache, er
...
 	if config.CleanWindow > 0 {
 		go func() {
+			logger := newLogger(config.Logger)
 			ticker := time.NewTicker(config.CleanWindow)
 			defer ticker.Stop()
 			for {
 				select {
 				case <-ctx.Done():
-					fmt.Println("ctx done, shutting down bigcache cleanup routine")
+					logger.Printf("ctx done, shutting down bigcache cleanup routine")
```
janisz, line 109:
```suggestion
					config.Logger.Printf("ctx done, shutting down bigcache cleanup routine")
```

bigcache.go
```diff
@@ -101,12 +100,13 @@ func newBigCache(ctx context.Context, config Config, clock clock) (*BigCache, er
 
 	if config.CleanWindow > 0 {
 		go func() {
+			logger := newLogger(config.Logger)
```
janisz, line 103: We don't nee to create new instance. Let's use `config.Logger.` inline
```suggestion
```

janisz (changes requested):

cristaloleg:
AFAIR it's because during a test run `os.Stdout` is replaced with a test-specific writer to catch the full output. And because this we're logging in a goroutine we can log after a test end, which results in a panic due to Go test cleanup.

If you're curious to fix it than try to replace this context with `context.WithCancel(...)` https://github.com/allegro/bigcache/blob/157556c5afaeebd739cdbff4755e65f88a5d95af/examples_test.go#L63

And if you're not curious (which is fine too) I will take a look later today (leaving PR open for some time).

Thanks.

bigcache.go
janisz, line 103: This will solve the data race
```suggestion
		logger := newLogger(config.Logger)
		go func() {
```

cristaloleg: Nice! anyway, replacing ctx is also a good thing for the future.

janisz:
Closing as this was fixed by:
- https://github.com/allegro/bigcache/pull/392

PR #377, opened 2023-10-14
changed to uint64 keys

### LINKED ISSUE #354: The index(int) of entries(queue.BytesQueue) overflows the hashmap's uint32 value
When the entries of a cacheShard is large enough, `set` a new item into the entries, then `get` the item could fail with "not found" error.

The reason is when the entries of a cacheShard is large enough, set a new item into the entries could result in a large index, then this index is converted to uint32 since the hashmap takes uint32 as the value.  

```
type cacheShard struct {
	hashmap     map[uint64]uint32
	entries     queue.BytesQueue
	lock        sync.RWMutex
	entryBuffer []byte
	onRemove    onRemoveCallback

	isVerbose    bool
	statsEnabled bool
	logger       Logger
	clock        clock
	lifeWindow   uint64

	hashmapStats map[uint64]uint32
	stats        Stats
}
```

```
if index, err := s.entries.Push(w); err == nil {
			s.hashmap[hashedKey] = uint32(index)
			s.lock.Unlock()
			return nil
		}
```

This issue is found with a stress test when I set an item to the cacheShard, I got an index value "5576813569", but when this index was added to the hashmap, it's converted to uint32, which is "1281846273". Therefore, when I tried to get this item, I got this wrong index(1281846273) from the hashmap, then I got the wrong entry.  Since the entry is wrong, the entryKey doesn't match with the actual key passed by `get` function, then BigCache considered this as a Collision and returned ErrEntryNotFound in the following code:
```
func (s *cacheShard) get(key string, hashedKey uint64) ([]byte, error) {
	s.lock.RLock()
	wrappedEntry, err := s.getWrappedEntry(hashedKey)
	if err != nil {
		s.lock.RUnlock()
		return nil, err
	}
	if entryKey := readKeyFromEntry(wrappedEntry); key != entryKey {
		s.lock.RUnlock()
		s.collision()
		if s.isVerbose {
			s.logger.Printf("Collision detected. Both %q and %q have the same hash %x", key, entryKey, hashedKey)
		}
		return nil, ErrEntryNotFound
	}
	entry := readEntry(wrappedEntry)
	s.lock.RUnlock()
	s.hit(hashedKey)

	return entry, nil
}
```

For more information on how to provide an MCVE, please see the [Stack Overflow documentation](https://stackoverflow.com/help/mcve).

**Environment:**
- Version (git sha or release): v3
- OS (e.g. from `/etc/os-release` or winver.exe): both Mac and Linux 64 bits
- go version: `go version go1.19.4 darwin/arm64`

raeidish:
closes  #354 

was trying to run benchmarks however when using uint32 keys i consistently ran into slice bounds out of range errors as seen below:

`BenchmarkAppendToCache/1-shards-12       panic: runtime error: slice bounds out of range [98993:98992]`

### DIFF
--- a/shard.go
+++ b/shard.go
@@ -16,7 +16,7 @@ type Metadata struct {
 }
 
 type cacheShard struct {
-	hashmap     map[uint64]uint32
+	hashmap     map[uint64]uint64
 	entries     queue.BytesQueue
 	lock        sync.RWMutex
 	entryBuffer []byte
@@ -140,7 +140,7 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
 
 	for {
 		if index, err := s.entries.Push(w); err == nil {
-			s.hashmap[hashedKey] = uint32(index)
+			s.hashmap[hashedKey] = uint64(index)
 			s.lock.Unlock()
 			return nil
 		}
@@ -164,7 +164,7 @@ func (s *cacheShard) addNewWithoutLock(key string, hashedKey uint64, entry []byt
 
 	for {
 		if index, err := s.entries.Push(w); err == nil {
-			s.hashmap[hashedKey] = uint32(index)
+			s.hashmap[hashedKey] = uint64(index)
 			return nil
 		}
 		if s.removeOldestEntry(NoSpace) != nil {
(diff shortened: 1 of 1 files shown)


cristaloleg (approved):
SGTM

cristaloleg:
@janisz ?

janisz (approved):

PR #382, opened 2023-10-28
feat: Add function to check if cache contains a key

Darkheir:
Add a new function allowing to check if the cache contains the given key


janisz:
Do we need this func? We have `Get`

cristaloleg (dismissed):
SGTM

shard.go
```diff
@@ -117,6 +117,16 @@ func (s *cacheShard) getValidWrapEntry(key string, hashedKey uint64) ([]byte, er
 	return wrappedEntry, nil
 }
 
+func (s *cacheShard) contains(key string, hashedKey uint64) bool {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
```
janisz, line 122: defer might have negative impact so we decided to not use it:
- https://github.com/allegro/bigcache/pull/37
- https://github.com/allegro/bigcache/pull/46
although things has changed so maybe it will not have a negative impact in general but in this case we can release lock before reading key from entry

janisz (changes requested):
Since @cristaloleg is fine with extending the interface let's merge it. Just one nitpic

shard.go
Darkheir: Done.

I thought the code was not using `defer` because it was made before the go 1.14 optimisations on defer statements, but the code is simple enough to not rely on it :-)

shard.go
```diff
@@ -117,6 +117,18 @@ func (s *cacheShard) getValidWrapEntry(key string, hashedKey uint64) ([]byte, er
...
 
+func (s *cacheShard) contains(key string, hashedKey uint64) bool {
+	s.lock.RLock()
+	wrappedEntry, err := s.getWrappedEntry(hashedKey)
+	if err != nil {
+		s.lock.RUnlock()
+		return false
+	}
+	entryKey := readKeyFromEntry(wrappedEntry)
+	s.lock.RUnlock()
```
janisz, line 128: @cristaloleg do you think we should increment hit count here?

janisz (dismissed):

shard.go
cristaloleg: Deeply philosophical question. I'm fine to not increment it here. Leave it only for real value get.

cristaloleg (dismissed):

PR #392, opened 2024-02-22
Only print if Verbose is true

xyproto:


bigcache.go
```diff
@@ -111,7 +111,9 @@ func newBigCache(ctx context.Context, config Config, clock clock) (*BigCache, er
 			for {
 				select {
 				case <-ctx.Done():
-					fmt.Println("ctx done, shutting down bigcache cleanup routine")
```
cristaloleg, line 114: TBH this can be absolutely dropped, WDYT @janisz ?

janisz: I agree

bigcache.go
```diff
@@ -111,7 +111,9 @@ func newBigCache(ctx context.Context, config Config, clock clock) (*BigCache, er
...
+					if config.Verbose {
+						fmt.Println("ctx done, shutting down bigcache cleanup routine")
+					}
```
janisz, line 116:
```suggestion
```

cristaloleg (approved):

xyproto:
Cool, thanks for reviewing and merging. 👍

PR #393, opened 2024-02-29
Fixing grammar in docs 

SaschaPeukert:
This is mainly about the difference between `bytes store` and `bytes stored` (which I would guess is the correct thing)

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -188,7 +188,7 @@ func (c *BigCache) ResetStats() error {
 	return nil
 }
 
-// Len computes the number of entries in the cache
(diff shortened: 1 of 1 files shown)


bigcache.go
```diff
@@ -188,7 +188,7 @@ func (c *BigCache) ResetStats() error {
 	return nil
 }
 
-// Len computes number of entries in cache
+// Len computes the number of entries in the cache
```
cristaloleg, line 191: Please, add a dot at the end.

SaschaPeukert: Done

cristaloleg (approved):

PR #396, opened 2024-06-27
feat: use mmap to alloc big chunk of memory

tecty:
Use mmap to malloc memory, give a smaller heap memory footprint. This will remove the go memory ballast effect while using bigcache. Thus this will increase a bit on GC stop and CPU. But we have other tools like GC params or gctuner to mitigate this kind of problem. 
image


README.md
```diff
@@ -1,4 +1,4 @@
-# BigCache [![Build Status](https://github.com/allegro/bigcache/workflows/build/badge.svg)](https://github.com/allegro/bigcache/actions?query=workflow%3Abuild)&nbsp;[![Coverage Status](https://coveralls.io/repos/github/allegro/bigcache/badge.svg?branch=main)](https://coveralls.io/github/allegro/bigcache?branch=main)&nbsp;[![GoDoc](https://godoc.org/github.com/allegro/bigcache/v3?status.svg)](https://godoc.org/github.com/allegro/bigcache/v3)&nbsp;[![Go Report Card](https://goreportcard.com/badge/github.com/allegro/bigcache/v3)](https://goreportcard.com/report/github.com/allegro/bigcache/v3)
+# BigCache [![Build Status](https://github.com/allegro/bigcache/workflows/build/badge.svg)](https://github.com/allegro/bigcache/actions?query=workflow%3Abuild)&nbsp;[![Coverage Status](https://coveralls.io/repos/github/allegro/bigcache/badge.svg?branch=main)](https://coveralls.io/github/allegro/bigcache?branch=main)&nbsp;[![GoDoc](https://godoc.org/github.com/tecty/bigcache/v3?status.svg)](https://godoc.org/github.com/tecty/bigcache/v3)&nbsp;[![Go Report Card](https://goreportcard.com/badge/github.com/tecty/bigcache/v3)](https://goreportcard.com/report/github.com/tecty/bigcache/v3)
```
janisz, line 1: Nice try :)

queue/bytes_queue.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/binary"
 	"log"
 	"time"
+
+	"golang.org/x/sys/unix"
```
janisz, line 8: I like that bigcache has no external deps
Also with unix we should add a build tags to not use it on Windows

README.md
tecty: I will use another branch to minimize the changes.

queue/bytes_queue.go
tecty: fixed on #397

Issue #318, opened 2022-05-22
the different Expired judge in cacheShard.getWithInfo() and cacheShard.onEvict()


chen-zhuohan:
**What is the issue you are having?**

the different Expired judge in cacheShard.getWithInfo() and cacheShard.onEvict()

```go
// github.com/allegro/bigcache/v3@v3.0.2/shard.go
func (s *cacheShard) onEvict(oldestEntry []byte, currentTimestamp uint64, evict func(reason RemoveReason) error) bool {
	...
	if currentTimestamp-oldestTimestamp > s.lifeWindow { // here, is larger
		evict(Expired)
		return true
	}
	return false
}

func (s *cacheShard) getWithInfo(key string, hashedKey uint64) (entry []byte, resp Response, err error) {
        ...
	if currentTime-oldestTimeStamp >= s.lifeWindow { // here, is larger or equal
		resp.EntryStatus = Expired
	}
}
```

**What is BigCache doing that it shouldn't?**

They should have the same judge.

**Environment:**
- Version: github.com/allegro/bigcache/v3@v3.0.2/shard.go

janisz:
Good catch! This inconsistency was introduced in https://github.com/allegro/bigcache/pull/168
I think we should have small helper method `isAlive` and use it instead of coping condition.

@chen-zhuohan   Would you like to contribute a patch?

```go
func isAlive() bool {
    return currentTimestamp-oldestTimestamp > s.lifeWindow
}```

chen-zhuohan:
#319 here

PR #403, opened 2024-08-07
Revert change to how expiration timings to calculated.

gurudesu:
When testing upgrading from [v3.0.2](https://github.com/allegro/bigcache/releases/tag/v3.0.2) to [v3.1.0](https://github.com/allegro/bigcache/releases/tag/v3.1.0), I've noticed a small bug impactful (to me) change around how expirations are calculated.

If you set a `LifeWindow` of `1s`, add an entry, and wait 1100ms, the entry does not expire. Testing it a bit more, I've discovered that it actually expires if you wait `2s`.

You can see this regression in code when comparing the old logic [here](https://github.com/allegro/bigcache/compare/v3.0.2...v3.1.0#diff-61374e089aecc0446ef16b857ce1b251d3231aa4dc9658c3d78c28f3525138fcL56) with the new logic [here](https://github.com/allegro/bigcache/compare/v3.0.2...v3.1.0#diff-61374e089aecc0446ef16b857ce1b251d3231aa4dc9658c3d78c28f3525138fcR289).

For convenience, old logic:

```golang
if currentTime-oldestTimeStamp >= s.lifeWindow {
}
```

new logic:

```golang
return currentTimestamp-oldestTimestamp > s.lifeWindow
```

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -256,7 +256,6 @@ func TestTimingEviction(t *testing.T) {
 	cache.Set("key", []byte("value"))
 
 	// when
-	clock.set(1)
 	cache.Set("key2", []byte("value2"))
 	_, err := cache.Get("key")
 
@@ -1181,7 +1180,7 @@ func TestBigCache_GetWithInfo(t *testing.T) {
 			name:     "Expired",
 			clock:    5,
 			wantData: value,
-			wantResp: Response{},
+			wantResp: Response{EntryStatus: Expired},
 		},
 		{
 			name:     "After Expired",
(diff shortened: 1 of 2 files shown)


bigcache_test.go
```diff
@@ -1181,7 +1180,7 @@ func TestBigCache_GetWithInfo(t *testing.T) {
 			name:     "Expired",
 			clock:    5,
 			wantData: value,
-			wantResp: Response{},
+			wantResp: Response{EntryStatus: Expired},
```
gurudesu, line 1183: This test case conflicts with the test name - if the name is `Expired` maybe it should be expected to be expired.

janisz:
I'm sorry for late answer. It was discussed here https://github.com/allegro/bigcache/issues/318
And yes, looks like we've choosen not the best option here :) 
Thanks for fixing, this

janisz (approved):

janisz:
@gurudesu For some reason CI has not run for this commit. Could you rebase and push again to trigger the CI?

Issue #400, opened 2024-07-13
Improve error handling in BytesQueue.Push function by using descriptive error messages in queue.go


akanchha25:
The Push function in BytesQueue currently uses a generic error message when the queue is full. It is suggested to use errFullQueue for a more descriptive error message.

Current Code:
```go
var (
	errEmptyQueue       = &queueError{"Empty queue"}
	errInvalidIndex     = &queueError{"Index must be greater than zero. Invalid index."}
	errIndexOutOfBounds = &queueError{"Index out of range"}
)
```

Push function
```go
if !q.canInsertAfterTail(neededSize) {
		if q.canInsertBeforeHead(neededSize) {
			q.tail = leftMarginIndex
		} else if q.capacity+neededSize >= q.maxCapacity && q.maxCapacity > 0 {
			return -1, &queueError{"Full queue. Maximum size limit reached."}
		} else {
			q.allocateAdditionalMemory(neededSize)
		}
	}
```

Updated code
```go
var (
	errEmptyQueue       = &queueError{"Queue is empty."}
	errInvalidIndex     = &queueError{"Index must be greater than zero. Invalid index."}
	errIndexOutOfBounds = &queueError{"Index out of range."}
	errFullQueue        = &queueError{"Queue is full. Maximum size limit reached."} 
)
```
```go
if !q.canInsertAfterTail(neededSize) {
		if q.canInsertBeforeHead(neededSize)) {
			q.tail = leftMarginIndex
		} else if q.capacity+neededSize >= q.maxCapacity && q.maxCapacity > 0 {
			return -1, errFullQueue 
		} else {
			q.allocateAdditionalMemory(neededSize)
		}
	}
```

This will improve the code quality

janisz:
@akanchha25 Indeed it's a good improvement. Do you think errors should be unexpeorted? If we export them then they could be used to check for given error like with `io.EOF`

akanchha25:
@janisz That's a great point. Exporting the errors would indeed allow for more robust error handling, similar to checking for specific errors like io.EOF. This could be beneficial for users of the BytesQueue package to handle different error scenarios more explicitly.

I can update the code to export these error variables, making it easier for others to handle specific error conditions. I'll proceed with the changes if that sounds good to you.

Vikram222726:
[...]

PR #405, opened 2025-01-14
Improve error handling in BytesQueue.Push function in queue.go

alonohana627:
Per [this issue](https://github.com/allegro/bigcache/issues/400). Added new error type and adjusted the tests accordingly.

### DIFF
--- a/queue/bytes_queue.go
+++ b/queue/bytes_queue.go
@@ -14,9 +14,10 @@ const (
 )
 
 var (
-	errEmptyQueue       = &queueError{"Empty queue"}
+	errEmptyQueue       = &queueError{"Queue is empty."}
 	errInvalidIndex     = &queueError{"Index must be greater than zero. Invalid index."}
 	errIndexOutOfBounds = &queueError{"Index out of range"}
+	errFullQueue        = &queueError{"Queue is full. Maximum size limit reached."}
 )
 
 // BytesQueue is a non-thread safe queue type of fifo based on bytes array.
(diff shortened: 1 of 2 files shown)


queue/bytes_queue.go
```diff
@@ -14,9 +14,10 @@ const (
 )
 
 var (
-	errEmptyQueue       = &queueError{"Empty queue"}
```
janisz, line 17: I prefer to not change it. Since it's part of public message, changes may break backward compatibility if someone is checking the error string.

queue/bytes_queue.go
```diff
@@ -14,9 +14,10 @@ const (
...
+	errEmptyQueue       = &queueError{"Queue is empty."}
 	errInvalidIndex     = &queueError{"Index must be greater than zero. Invalid index."}
 	errIndexOutOfBounds = &queueError{"Index out of range"}
+	errFullQueue        = &queueError{"Queue is full. Maximum size limit reached."}
```
janisz, line 20: :+1:

queue/bytes_queue.go
alonohana627: Reverted. Done.

queue/bytes_queue.go
janisz:
```suggestion
	errFullQueue        = &queueError{"Full queue. Maximum size limit reached."}
```

backward compatibility

queue/bytes_queue_test.go
```diff
@@ -399,7 +399,7 @@ func TestMaxSizeLimit(t *testing.T) {
 
 	// then
 	assertEqual(t, 50, capacity)
-	assertEqual(t, "Full queue. Maximum size limit reached.", err.Error())
+	assertEqual(t, "Queue is full. Maximum size limit reached.", err.Error())
```
janisz, line 402:
```suggestion
	assertEqual(t, "Full queue. Maximum size limit reached.", err.Error())
```

PR #414, opened 2025-09-10
Add ResetKeyMetadata method to resets per-key metadata.

askomorokhov:
Add `ResetKeyMetadata` Method to Clear Per-Key Request Counts
This PR introduces a `ResetKeyMetadata` method to `BigCache`, allowing users to reset per-key metadata such as request counts.
Similar to the existing `Reset` and `ResetStats` methods, `ResetKeyMetadata` helps clean up accumulated request statistics without affecting cached entries.

This is useful for applications that need to periodically clear usage metrics while retaining cache data.

**Summary of changes:**

- Added `BigCache.ResetKeyMetadata()` and `cacheShard.resetKeyMetadata()` to clear per-key request counts.
- Ensures thread safety by locking shards during reset.

### DIFF
--- a/bigcache.go
+++ b/bigcache.go
@@ -188,6 +188,14 @@ func (c *BigCache) ResetStats() error {
 	return nil
 }
 
+// ResetKeyMetadata resets per-key metadata
+func (c *BigCache) ResetKeyMetadata() error {
+	for _, shard := range c.shards {
+		shard.resetKeyMetadata(c.config)
+	}
+	return nil
+}
+
 // Len computes the number of entries in the cache.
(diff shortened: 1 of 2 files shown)


bigcache.go
```diff
@@ -188,6 +188,14 @@ func (c *BigCache) ResetStats() error {
 	return nil
 }
 
+// ResetKeyMetadata resets per-key metadata
+func (c *BigCache) ResetKeyMetadata() error {
```
janisz, line 192:
```suggestion
func (c *BigCache) ResetKeysMetadata() error {
```

Issue #419, opened 2026-02-04
Optimize map capacity allocation during shard initialization.


xiaobai12011:
If StatsEnabled is configured as false, could we consider not allocating a ShardSize size for hashmapStats during the Shard initialization function (initNewShard)? This is because the data is not used in subsequent processes, resulting in the allocation of invalid memory space. This issue is more pronounced when caching a large number of entries.
```
func initNewShard(config Config, callback onRemoveCallback, clock clock) *cacheShard {
	bytesQueueInitialCapacity := config.initialShardSize() * config.MaxEntrySize
	maximumShardSizeInBytes := config.maximumShardSizeInBytes()
	if maximumShardSizeInBytes > 0 && bytesQueueInitialCapacity > maximumShardSizeInBytes {
		bytesQueueInitialCapacity = maximumShardSizeInBytes
	}
	return &cacheShard{
		hashmap:      make(map[uint64]uint64, config.initialShardSize()),
                // if StatsEnabled==false，Only  make(map[uint64]uint32)
		hashmapStats: make(map[uint64]uint32, config.initialShardSize()),
		entries:      *queue.NewBytesQueue(bytesQueueInitialCapacity, maximumShardSizeInBytes, config.Verbose),
		entryBuffer:  make([]byte, config.MaxEntrySize+headersSizeInBytes),
		onRemove:     callback,
		isVerbose:    config.Verbose,
		logger:       newLogger(config.Logger),
		clock:        clock,
		lifeWindow:   uint64(config.LifeWindow.Seconds()),
		statsEnabled: config.StatsEnabled,
		cleanEnabled: config.CleanWindow > 0,
	}
}
```

janisz:
Sounds good. Would you like to provide a patch?

xiaobai12011:
> Sounds good. Would you like to provide a patch?

https://github.com/allegro/bigcache/pull/420/changes

ok

PR #420, opened 2026-02-05
Optimize map capacity allocation during shard initialization. #419

xiaobai12011:
If StatsEnabled is configured as false, No need to initialize the capacity for hashmapStats.
This is because the data is not used in subsequent processes, resulting in the allocation of invalid memory space.


janisz (approved):

janisz:
One test is failing

xiaobai12011:
> One test is failing

I ran the TestCacheReset test locally and didn't encounter any failures. Could you please help me troubleshoot or rerun the test?

PR #421, opened 2026-04-21
fix(shard): defer shard lock unlock so panics inside eviction do not strand writers

### LINKED ISSUE #401: A panic occurred, causing the lock to remain engaged.panic: makeslice: len out of range
A panic occurred, causing the lock to remain engaged. Even if I use recover from the panic, the program cannot continue running because the lock is not released. When other goroutines try to use bigcache, they cannot acquire the lock and will wait indefinitely.

traces:
/data/bkdevops/apps/golang/1.18/src/runtime/panic-go: 838 +0x207
github.com/allegro/bigcache/v2.readEntry(...)
/root/go/pkg/mod/github.com/allegro/bigcache/v2@v2.2.5/encoding-go:51
github. com/allegro/bigcache/v2. (*BigCache). providedOnRemoveWithReason(0xc55c947450, (0xc6b528cbbc, 0x34, 0X1145444}, 0x1)
/root/go/pkg/mod/github.com/allegro/bigcache/v2@v2.2.5/bigcache.go:229 +0x1ae
github.com/allegro/bigcache/v2. (*cacheShard). remove0ldestEntry(0xc5ae07e360, 0x2c73c338?)
/root/go/pkg/mod/github.com/allegro/bigcache/v2@v2.2.5/shard-go:324 +0xc9
github.com/allegro/bigcache/v2. (*cacheShard).onEvict(0x1186180?, {0xcbb528cbbc?,
0x27a5bcd3a24d8539?, 0x0?}, x0?, 0xc53c1f73e0?)
/root/go/pkg/mod/github.com/allegro/bigcache/v2@v2.2.5/shard.go:271 +0x3f
github.com/allegro/bigcache/v2. (*cacheShard).set(0xc5ae07e360, {0xcb043126c0, 0x5a}, 0x27a5bcd3a24d8539, {0xc48f9e53c8, 0x20, 0xc48f9
e53c8?})
/root/go/pkg/mod/github.com/allegro/bigcache/v2@v2.2.5/shard.go:132 +0x178
github.com/allegro/bigcache/v2. (*BigCache).Set(0xc55c947450, {0xcb043126c0, 0x5a}, {0xc48f9e53c8, 0x20, 0x20})
/root/go/pkg/mod/github.com/allegro/bigcache/v2@v2.2.5/bigcache.go:133 +0x8d

bigcache config:
Shards:             512,             
LifeWindow:         60 * time.Minute,
CleanWindow:        time.Second * 60, 
MaxEntrySize:       4096,
Verbose:            false,
HardMaxCacheSize:   64 * 1024,

other info:
i put key to bigcache 50000/min
60000 keys in bigcache 

i can`t provide minimal,because I cann`t reliably reproduce this problem. It occurs randomly within 2 to 48 hours after the program starts.

**Environment:**
- Version (git sha or release): v2.2.5
- OS (e.g. from `/etc/os-release` or winver.exe): linux
- go version:  1.18

SAY-5:
## What

Fixes #401.

`cacheShard.set` and `cacheShard.append` acquired `s.lock` with explicit `s.lock.Unlock()` calls along each return path. Any panic fired between the `Lock()` and the first `Unlock()` - for example the `makeslice: len out of range` the user hit inside `readEntry` / `providedOnRemoveWithReason` during an `onEvict` on a corrupted entry - unwound through the deferred nothing and left the shard permanently write-locked. Every subsequent `Set` / `Append` / `Delete` on that shard's key space then blocked forever while the rest of the app kept running, producing the hang that manifests as 'cache seems deadlocked' in production.

## Fix

Acquire the lock at the top of each function and release it with `defer` so that a panic inside the eviction / serialization path is still caught by any higher-level recover, but the shard is always unlocked as the goroutine unwinds. Removed the redundant manual `Unlock` calls along the existing return sites.

Runtime hot-path is functionally unchanged: the defer adds one mov/call per `Set` compared to zero, negligible next to the map + queue work on either side.

## Verification

Locally on macOS, go 1.26.2:
- `gofmt -s -l shard.go`: clean
- `go vet ./...`: clean
- `go test -race -count=1 -short -run 'TestCacheSet|TestCacheGet|TestCacheLen|TestCacheCapacity|TestCacheInitialCapacity|TestCacheDel' ./...`: pass

Note on `TestCacheReset`: this test fails on unmodified master as well (`expected: int(1337) actual: int(725)`) and looks pre-existing / flaky - definitely not a regression from this change. Happy to investigate separately if desired.

Closes #401

### DIFF
--- a/shard.go
+++ b/shard.go
@@ -121,6 +121,14 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
 	currentTimestamp := uint64(s.clock.Epoch())
 
 	s.lock.Lock()
+	// Hold the shard lock with defer so that a panic deeper in
+	// readEntry / onRemove / providedOnRemoveWithReason (e.g. a
+	// makeslice-len-out-of-range from a corrupted entry) does not
+	// leave the lock permanently held. Previously, a panic during
+	// onEvict would tear down the goroutine before the manual
+	// Unlock below could run, and every subsequent writer on the
+	// same shard blocked forever (#401).
+	defer s.lock.Unlock()
 
 	if previousIndex := s.hashmap[hashedKey]; previousIndex != 0 {
 		if previousEntry, err := s.entries.Get(int(previousIndex)); err == nil {
@@ -141,11 +149,9 @@ func (s *cacheShard) set(key string, hashedKey uint64, entry []byte) error {
 	for {
 		if index, err := s.entries.Push(w); err == nil {
 			s.hashmap[hashedKey] = uint64(index)
-			s.lock.Unlock()
 			return nil
 		}
 		if s.removeOldestEntry(NoSpace) != nil {
-			s.lock.Unlock()
 			return errors.New("entry is bigger than max shard size")
 		}
 	}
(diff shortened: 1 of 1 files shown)


janisz (commented):
Unfortunately it does not solves the panic, it only allows to handle that gracefully.

PR #426, opened 2026-07-21
fix(shard): defer unlock so onRemove panics cannot strand writers

### LINKED ISSUE #401: A panic occurred, causing the lock to remain engaged.panic: makeslice: len out of range

Solaris-star:
## Summary
When `OnRemove` / `OnRemoveWithReason` panics under a shard write lock (issue #401: `makeslice: len out of range` inside `readEntry` during eviction), the old manual `Unlock()` paths never ran. That left the shard mutex held forever and hung every subsequent writer, even after `recover` in the caller.

## Fix
Use `defer s.lock.Unlock()` on the write paths that call into remove/onRemove callbacks (`set`, `append`, `del`, `cleanUp`, `reset`, `resetStats`).

## Test plan
- [x] `go test .`
- [x] `TestShardUnlocksOnOnRemovePanic` — panic in `OnRemoveWithReason` during `Delete`, then assert a later `Set` does not hang

Fixes #401

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -1388,3 +1388,55 @@ func TestRemoveNonExpiredData(t *testing.T) {
 		noError(t, err)
 	}
 }
+
+
+func TestShardUnlocksOnOnRemovePanic(t *testing.T) {
+	// A panic inside onRemove used to leave the shard mutex locked forever
+	// because set/del unlocked manually after the callback (#401).
+	// After the fix, defer unlocks and subsequent writes can proceed.
+	cache, err := New(context.Background(), Config{
+		Shards:             1,
+		LifeWindow:         time.Minute,
+		CleanWindow:        0,
+		MaxEntriesInWindow: 10,
+		MaxEntrySize:       64,
+		OnRemoveWithReason: func(key string, entry []byte, reason RemoveReason) {
+			panic("simulated onRemove panic")
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := cache.Set("key", []byte("value")); err != nil {
+		t.Fatal(err)
+	}
+
+	panicked := false
+	func() {
+		defer func() {
+			if recover() != nil {
+				panicked = true
+			}
+		}()
+		_ = cache.Delete("key")
+	}()
+	if !panicked {
+		t.Fatal("expected Delete to panic via OnRemoveWithReason")
+	}
+
+	done := make(chan error, 1)
+	go func() {
+		done <- cache.Set("after-panic", []byte("ok"))
+	}()
+
+	select {
+	case err := <-done:
+		if err != nil {
+			t.Fatalf("Set after panic failed: %v", err)
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatal("shard still locked after onRemove panic")
+	}
+}
+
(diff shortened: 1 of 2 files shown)


janisz:
I'm not sure if that's a good solution to the panic problem. We should fix panic not mask it (or allow handling) by `defer`. Did you find a way to reproduce panic?

janisz:
I think I found a root cause of the panic:
- https://github.com/allegro/bigcache/pull/427

Solaris-star:
You're right — this masks the panic rather than fixing it, and I'll be honest: I didn't reproduce the root cause, I only addressed the lock-stranding consequence. Deferring to #427, which fixes the actual root cause (peek() missing bounds validation on a stale index pointing at corrupt uvarint data). Closing this in favor of that. Happy to help review #427 if useful.

Issue #338, opened 2022-10-06
bytesqueue peak panic


tcsgogogo:
found a panic:
runtime error: slice bounds out of range [:31874] with capacity 24992
goroutine 1708235262 [running]:
panic({0x1618300, 0xc030365740})
        /data/golang/1.18/src/runtime/panic.go:838 +0x207
github.com/allegro/bigcache/v3/queue.(*BytesQueue).peek(0x140e960?, 0xc001730630?)
        /root/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.2/queue/bytes_queue.go:242 +0x177
github.com/allegro/bigcache/v3/queue.(*BytesQueue).Get(...)
        /root/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.2/queue/bytes_queue.go:193
github.com/allegro/bigcache/v3.(*cacheShard).getWrappedEntry(0xc0004f5200, 0xc06ab11d60?)
        /root/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.2/shard.go:92 +0x46
github.com/allegro/bigcache/v3.(*cacheShard).get(0xc0004f5200, {0x18a8321, 0x1}, 0x1000000000001?)
        /root/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.2/shard.go:64 +0x7e
github.com/allegro/bigcache/v3.(*BigCache).Get(0xc002c324e0, {0x18a8321, 0x1})
        /root/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.2/bigcache.go:123 +0x69

**Environment:**
- Version (git sha or release): v3.0.2
- OS (e.g. from `/etc/os-release` or winver.exe): centos 
- go version:  1.18

janisz:
Do you have any way to reproduce it?

janisz:
Interesting, what entries and keys looks like? Were the keys randomly distributed?

tcsgogogo:
> Interesting, what entries and keys looks like? Were the keys randomly distributed?

1) In fact, both key and value are uint64 numbers.
2) The key comes from the uint64 number entered by the user, so it may not be randomly distributed.

Will data skew in a shard cause this problem? Shouldn't it cause a set error rather than a get error, let alone a panic.

janisz:
> Will data skew in a shard cause this problem? Shouldn't it cause a set error rather than a get error, let alone a panic.

[...]

Issue #312, opened 2022-03-25
shard onEvict get an empty oldestEntry.so panic and not unLock


s1040735149:
**What is the issue you are having?**

shard onEvict get an empty oldestEntry.so panic and not unLock

**What is BigCache doing that it shouldn't?**
fix panic

**Minimal, Complete, and Verifiable Example**

runtime error: index out of range [7] with length 0
goroutine 60116345 [running]:
runtime/debug.Stack(0x2153be0, 0x22c8700, 0xc1ec77bf40)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9f
git.in.zhihu.com/go/utils.SafelyRun.func1(0xc1ea5fdf40)
	/go/pkg/mod/git.in.zhihu.com/go/utils@v0.2.15/concurrent.go:20 +0x78
panic(0x22c8700, 0xc1ec77bf40)
	/usr/local/go/src/runtime/panic.go:969 +0x175
encoding/binary.littleEndian.Uint64(...)
	/usr/local/go/src/encoding/binary/binary.go:77
github.com/allegro/bigcache/v3.readTimestampFromEntry(...)
	/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.1/encoding.go:58
github.com/allegro/bigcache/v3.(*cacheShard).onEvict(0xc0ba3fd7a0, 0xc166605820, 0x0, 0xbc7e0, 0x623d7f7d, 0xc1ea5fd730, 0x0)
	/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.1/shard.go:271 +0x85
github.com/allegro/bigcache/v3.(*cacheShard).set(0xc0ba3fd7a0, 0xc1ecdc1ce0, 0x23, 0xd8bc05fe228582f8, 0xc1ecdd9200, 0x29d, 0x2c6, 0x0, 0x0)
	/go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.1/shard.go:134 +0x31a
github.com/allegro/bigcache/v3.(*BigCache).Set(0xc000214680, 0xc1ecdc1ce0, 0x23, 0xc1ecdd9200, 0x29d, 0x2c6, 0x0, 0x2105a00)

When asking a question about a problem caused by your code, you will get much better answers if you provide code we can use to reproduce the problem. That code should be...

* ...Minimal – Use as little code as possible that still produces the same problem
* ...Complete – Provide all parts needed to reproduce the problem
[...]

Issue #305, opened 2022-01-12
The problem still exists in version 3.0.1，panic: runtime error: index out of range [7] with length 0 


jesse-engineer:
> In theory, this issues is fix in v3.0.1, but I still met
> 
> ```
> runtime error: index out of range [7] with length 0
> 
> runtime/debug.Stack()
>         /home/xx/Apps/go/src/runtime/debug/stack.go:24 +0x65
> golang.org/x/sync/singleflight.newPanicError({0x120c9a0, 0xc26e4c02d0})
>         /home/xx/.go/pkg/mod/golang.org/x/sync@v0.0.0-20201207232520-09787c993a3a/singleflight/singleflight.go:35 +0x2c
> golang.org/x/sync/singleflight.(*Group).doCall.func2.1()
>         /home/xx/.go/pkg/mod/golang.org/x/sync@v0.0.0-20201207232520-09787c993a3a/singleflight/singleflight.go:188 +0x3b
> panic({0x120c9a0, 0xc26e4c02d0})
>         /home/xx/Apps/go/src/runtime/panic.go:1038 +0x215
> encoding/binary.littleEndian.Uint64(...)
>         /home/xx/Apps/go/src/encoding/binary/binary.go:77
> github.com/allegro/bigcache/v3.readTimestampFromEntry(...)
>         /home/xx/.go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.1/encoding.go:58
> github.com/allegro/bigcache/v3.(*cacheShard).onEvict(0xc0edece368, {0xc377a92025, 0x50beccc039db3211, 0x0}, 0x0, 0x0)
>         /home/xx/.go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.1/shard.go:271 +0x65
> github.com/allegro/bigcache/v3.(*cacheShard).set(0xc0edece360, {0xc2326b0550, 0x4d}, 0x0, {0xc0dc931800, 0x7e, 0x400})
>         /home/xx/.go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.1/shard.go:134 +0x1a5
> github.com/xx/bigcache/v3.(*BigCache).Set(0xc0001b61a0, {0xc2326b0550, 0x4d}, {0xc0dc931800, 0x7e, 0x400})
>         /home/xx/.go/pkg/mod/github.com/allegro/bigcache/v3@v3.0.1/bigcache.go:133 +0x8d
[...]

PR #427, opened 2026-07-29
Fix panic in peek() when stale index points to corrupt uvarint data

janisz:
## Summary

Fixes `panic: makeslice: len out of range` in `peek()` when a stale hashmap index points to overwritten queue data containing uvarint continuation bytes (`0x80`+).

## Root cause

`BytesQueue.peek()` calls `binary.Uvarint()` on the data at a given index and uses the decoded `blockSize` to slice the array — with **no validation**. When the queue wraps and overwrites old positions, a stale index from the shard's hashmap can point to arbitrary data. If that data happens to contain 10+ continuation bytes (`0xff`), `binary.Uvarint` returns `n = -10` (overflow), and the subsequent slice `array[idx+(-10) : idx+0]` panics with `slice bounds out of range`.

The original filler-blob off-by-one that created corrupt entries was fixed in:
- https://github.com/allegro/bigcache/pull/266

but `peek()` itself never gained bounds checking — any future source of corruption (race, bitflip, stale index) would still panic instead of returning an error.

## Fix

Three validation checks added to `peek()` in `queue/bytes_queue.go`, all in `uint64` space to avoid silent truncation on 32-bit platforms:

1. **`n <= 0`** — uvarint decode failed (overflow or empty buffer)
2. **`blockSize < uint64(n)`** — decoded size is smaller than the header itself (corrupt)
3. **`uint64(index) + blockSize > uint64(len(q.array))`** — would slice past the array

## Test plan

- **`TestGetStaleIndexReturnsError`** — deterministic reproduction of #401 through public API: pushes `0xff` bytes, wraps the queue, then reads a stale index. Panics on master, passes with fix.
- **`TestPushAndGetEmptyEntry`** — guards against `<=` vs `<` regression (zero-length entries are valid).
- **`FuzzBytesQueue`** — random Push/Pop/Get/Peek sequences on `BytesQueue`.
- **`FuzzBigCache`** — random Set/Get/Delete through the public `BigCache` API with a tiny cache config.

## Breaking changes

- `go.mod` bumped from `go 1.16` to `go 1.22` (required for fuzz test support and rand v2). Go 1.16 reached EOL in March 2022.

Directly fixed

  - #148
  - #235
  - #250
  - #338
  - #401

Partially fixed — peek() catches corrupt blockSize but not corrupt internal fields

  - #222
  - #226
  - #234
  - #254
  - #257
  - #283
  - #290
  - #305

Already fixed by #266 — peek() adds defense-in-depth

  - #258
  - #285
  - #297
  - #300
  - #312

Not fixed — different root cause

  - #205 (config overflow)
  - #208 (OOM)
  - #263 (push/copy panic)


cristaloleg (dismissed):
Nice 🚀

Solaris-star (commented):
Reviewed the core `peek()` fix in detail — the bounds validation is correct. Walking through the three checks against the slice `q.array[index+n : index+int(blockSize)]`:

1. **`n <= 0`** — `binary.Uvarint` returns `n=0` when the buffer is too small to hold a full uvarint, and `n<0` on 64-bit overflow. Catching both here is right; previously a negative `n` flowed straight into the slice expression.

2. **`blockSize < uint64(n)`** — since `blockSize` is the *total* block size including the uvarint header, the payload slice `[index+n : index+blockSize]` inverts (start > end) when `blockSize < n`. This guard prevents that. Correct.

3. **`blockSize > uint64(len(q.array)-index)`** — bounds the slice end at `len(array)`. Correct.

Doing all comparisons in `uint64` space before the `int(blockSize)` conversion is the right call for 32-bit platforms — once check 3 passes, `blockSize <= len(q.array)-index` is guaranteed to fit in an `int`, so the conversion can't silently truncate.

This is the right layer to fix it too: even with the #266 filler-blob off-by-one resolved, `peek()` had no defense against any *other* source of a stale/corrupt index (race, bitflip), so returning `errIndexOutOfBounds` instead of panicking is proper defense-in-depth rather than masking.

One minor, non-blocking observation: the PR bundles the core fix with an unrelated `math/rand` → `math/rand/v2` migration and a new fuzz harness. The fuzz test is a great addition (it's exactly what would have surfaced this), but the rand modernization is orthogonal and makes the diff noisier than the fix warrants — splitting it would make the security-relevant change easier to bisect. Not a blocker.

janisz:
@Solaris-star splitted the PR and it's based on:
- #428

cristaloleg (approved):

PR #428, opened 2026-07-30
Bump Go to 1.22, migrate to math/rand/v2, fix staticcheck SA1006

janisz:
- Bump go.mod from 1.16 to 1.22
- Replace deprecated rand.Seed/rand.Read with math/rand/v2 and crypto/rand
- Fix staticcheck SA1006: use t.Errorf format directly instead of fmt.Sprintf

### DIFF
--- a/bigcache_test.go
+++ b/bigcache_test.go
@@ -842,7 +842,7 @@ func TestCacheReset(t *testing.T) {
 	// given
 	cache, _ := New(context.Background(), Config{
 		Shards:             8,
-		LifeWindow:         time.Second,
+		LifeWindow:         time.Minute,
 		MaxEntriesInWindow: 1,
 		MaxEntrySize:       256,
 	})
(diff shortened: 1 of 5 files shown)


bigcache_test.go
```diff
@@ -5,7 +5,8 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"math/rand"
+	crand "crypto/rand"
```
cristaloleg: Better to keep `"crypto/rand"` but `mrand "math/rand/v2"`.

cristaloleg (dismissed):
SGTM

janisz:
@cristaloleg PTAL

Aniruddhraam:
But can't this be bumped to the stable 1.26 directly instead. This seems like double the work to me. Anyways post 1.25 with the green tea GC if bigcache were to be accessed with the server, it would be significantly faster.

cristaloleg (approved):

cristaloleg:
> 1.26 directly instead

Go version in `go.mod` is just a requirement for the minimal Go version. If you build your project with Go 1.25+ you will get all benefits on newer runtime. No need to force the user.

PR #429, opened 2026-08-03
Fixed `go test -race -bench=. .` when running in parallel systems to prevent data races

Aniruddhraam:
### Summary
Fixes data races flagged when running `go test -race -bench=.` in `bigcache_bench_test.go`.

### Root Cause & Changes

1. **Moved `b.ReportAllocs()` outside `b.RunParallel`**
   - `b.ReportAllocs()` mutates non-thread-safe internal fields on `*testing.B`.
   - Calling `b.ReportAllocs()` inside `b.RunParallel(...)` caused concurrent write data races on `*testing.B` across parallel worker goroutines.
   - **Fix**: Relocated `b.ReportAllocs()` to the benchmark setup phase outside `b.RunParallel` across `writeToCache`, `appendToCache`, `readFromCache`, `readFromCacheNonExistentKeys`, and `BenchmarkIterateOverCache`.

2. **Isolated `EntryInfoIterator` instances per worker goroutine**
   - `EntryInfoIterator` is a stateful cursor. While `SetNext()` locks internal state during element advancement, it releases the lock before returning, allowing `Value()` to read state without holding the lock.
   - Sharing a single `Iterator` instance across multiple parallel goroutines inside `b.RunParallel` caused data races between concurrent `SetNext()` writes and `Value()` reads.
   - **Fix**: Moved `it := cache.Iterator()` inside the `b.RunParallel` loop so each worker goroutine operates on its own iterator instance, and added a documentation comment explaining iterator thread isolation.

### Verification
Ran benchmarks with the Go race detector enabled:
```bash
go test -race -bench=. .

### DIFF
--- a/bigcache_bench_test.go
+++ b/bigcache_bench_test.go
@@ -93,11 +93,13 @@ func BenchmarkIterateOverCache(b *testing.B) {
 				cache.Set(fmt.Sprintf("key-%d", i), m)
 			}
 
+			b.ReportAllocs()
 			b.ResetTimer()
-			it := cache.Iterator()
 
 			b.RunParallel(func(pb *testing.PB) {
-				b.ReportAllocs()
+				// EntryInfoIterator is a stateful cursor intended to be used by a single goroutine at a time.
+				// We instantiate a new iterator for each parallel goroutine to prevent data races.
+				it := cache.Iterator()
 
 				for pb.Next() {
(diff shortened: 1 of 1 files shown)


cristaloleg (approved):
SGTM

janisz (approved):

Issue #322, opened 2022-06-24
Reset() don't work


abolinhas:
**What is the issue you are having?**
Reset shards no longer works

**What is BigCache doing that it shouldn't?**
Empty the shards.

Using stats method i can see the how many items are in cache
```
cache, initErr := bigcache.NewBigCache(config)
cache.Stats()

Output: {"hits":1222,"misses":2,"delete_hits":0,"delete_misses":0,"collisions":0}
```
but when i run the method Reset to flush the shards, they are not flushed / empty, , in previously versions this method works correctly.

```
cache, initErr := bigcache.NewBigCache(config)
cache.Reset()
```
After i run the Reset() the stats are se same
`Output: {"hits":1222,"misses":2,"delete_hits":0,"delete_misses":0,"collisions":0}`

**Environment:**
- Version (git sha or release): v3@v3.0.2
- OS (e.g. from `/etc/os-release` or winver.exe):  Debian GNU/Linux 10 (buster)
- go version: 1.18.3

janisz:
Hi, I created following tests and it looks like stats were never cleared. What version do you mean in this sentence:
> in previously versions this method works correctly

```go
package bigcache

import (
	"fmt"
	"testing"
	"time"
)

func TestCacheReset322(t *testing.T) {
	t.Parallel()

	// given
	cache, _ := NewBigCache(Config{
		Shards:             8,
		LifeWindow:         time.Second,
		MaxEntriesInWindow: 1,
		MaxEntrySize:       256,
	})
	keys := 1337

	// when
	for i := 0; i < keys; i++ {
		cache.Set(fmt.Sprintf("key%d", i), []byte("value"))
		cache.Get(fmt.Sprintf("key%d", i))
	}

	// then
	if int64(keys) != cache.Stats().Hits {
		t.Error("Stats not set")
	}

	// and when
	cache.Reset()

	// then
	if int64(0) != cache.Stats().Hits {
		t.Error("Stats not cleared")
	}
}
```

Hojun-Cho:
Hello, I think this problem is because Reset() did not reset the cacheShard.Stats 
```go
func (c *BigCache) Reset() error {
	for _, shard := range c.shards {
		shard.reset(c.config)
	}
	return nil
}

func (s *cacheShard) reset(config Config) {
	s.lock.Lock()
	s.hashmap = make(map[uint64]uint32, config.initialShardSize())
[...]

PR #432, opened 2026-08-26
fix: clear per-key metadata (hashmapStats) on Reset

jaideeppyne:
`cacheShard.reset` rebuilds `hashmap`, `entryBuffer` and the `entries` queue but leaves `hashmapStats` untouched.

Everywhere else the invariant holds that `hashmapStats` only contains counts for currently-present entries — `del` and `removeOldestEntry` both `delete(s.hashmapStats, hash)` when an entry leaves the cache. `Reset()` removes every entry yet keeps their request counts, so:

1. after `Reset()`, `KeyMetadata(key).RequestCount` still reports a stale, non-zero value for keys that are no longer cached, and
2. the map grows without bound across `Reset` cycles when the key set churns (a leak).

This re-creates `hashmapStats` in `reset()`, mirroring `initNewShard` (zero capacity when `StatsEnabled` is false, so nothing is allocated in the disabled case).

Adds `TestResetClearsKeyMetadata`, which fails before this change (`RequestCount == 10` after `Reset`) and passes after (`== 0`). `go test ./...`, `-race`, `gofmt`, and `go vet` all pass.

Related: #322 (Reset not clearing stats), #414 (adds a separate `ResetKeyMetadata` API — orthogonal to this fix).


shard.go
```diff
@@ -346,8 +346,13 @@ func (s *cacheShard) removeOldestEntry(reason RemoveReason) error {
 }
 
 func (s *cacheShard) reset(config Config) {
+	var hashmapStatsCapacity int
+	if config.StatsEnabled {
+		hashmapStatsCapacity = config.initialShardSize()
+	}
 	s.lock.Lock()
 	s.hashmap = make(map[uint64]uint64, config.initialShardSize())
+	s.hashmapStats = make(map[uint64]uint32, hashmapStatsCapacity)
```
janisz, line 355: I wonder if we should use clear instead
```suggestion
	clear(s.hashmap)
	clear(s.hashmapStats)
```

jaideeppyne:
The red build here is `TestCacheLen`, which I don't think is related to this change.

It builds a cache with `LifeWindow: time.Second` and no `CleanWindow`, then writes 1337 keys and asserts `Len()` is 1337. It's also `t.Parallel()`, so on a loaded runner the write loop can outlast the one second life window, and the oldest entries are then expired and get evicted as later Sets grow the queue. That matches the failure, 933 instead of 1337.

This PR only touches `hashmapStats` in `reset()`, which isn't on that path. Locally the full suite passes on both main and this branch, and `TestCacheLen` on its own passes 12 out of 12 on each.

Happy to push an empty commit if you want CI re-run.

shard.go
jaideeppyne: `clear` does work here, and it releases the memory. I assumed it would keep the buckets, but that is wrong: filling a `map[uint64]uint32` with 2M entries and then calling `clear` takes the heap from 72.2 MB back to 0.2 MB, same as dropping the map, so there is no leak either way.

The one thing that changes is the size hint. `reset` takes `config` so it can build both maps at `initialShardSize()`, and a cleared map starts from nothing and grows again as the shard refills. There is also `hashmapStatsCapacity`, which is deliberately 0 when `StatsEnabled` is false so the stats map costs nothing in the common case.

So it is a tradeoff between one allocation at `Reset` and rehashing during the refill after it. I leaned towards keeping the hint since `reset` already has the config in hand, but I do not feel strongly, and switching both to `clear` is a smaller diff. Happy to push whichever you prefer.

jaideeppyne:
The red `Build (stable)` here is `TestCacheLen`, which I think is a pre-existing timing flake rather than anything in this change. I opened #433 with a deterministic repro on master.

This PR only touches `cacheShard.reset`, and `TestCacheLen` never calls `Reset`. `Build (oldstable)` passed on the same commit.

janisz (approved):

